PR #7, opened 2013-10-31
Move storing options on the element to after checks for methods like reload/destroy.

andyburke:
I was seeing a bug where I was calling 'destroy' before the sortable had been bound (which, in my opinion, should be a safe action). It was adding strange options to the element that would then be used when I tried to bind the element for the first time.

Moving the data storage to after the main methods seems to fix the issue.

### DIFF
--- a/jquery.sortable.js
+++ b/jquery.sortable.js
@@ -78,6 +78,9 @@ $.fn.sortable = function(options) {
       }
       dragging = null;
     }).not('a[href], img').on('selectstart.h5s', function() {
+      if (options.handle && !isHandle) {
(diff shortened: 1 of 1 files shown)


andyburke:
I also just fixed an issue where the click event on a div within a sortable element would be eaten, even if you're using handles.

voidberg:
Sorry, was very busy. Will merge this weekend. Thanks!

andyburke:
No problem, no rush.

voidberg:
Thanks and sorry for the delay. :)

PR #12, opened 2014-01-08
Added possibility for a sortable list to have a sortable child list

drskullster:
Currently, when applying sortable plugin on a child element of a list that is also sortable, parent element works as expected but child element does not.

What happens is that the events for the child element are propagated to the parent element. This pull request aims to fix that behaviour by stopping propagation on `dragstart` and on `drop`.

So you can do this :

```
<ul id="parent_list">
  <li>
    <ul id="child_list">
      <li></li>
    </ul>
  </li>
</ul>
<script type="text/javascript">
  $("#parent_list").sortable();
  $("#child_list").sortable();
</script>
```


voidberg:
Thanks!

PR #15, opened 2014-01-27
Detach dragged element from DOM while dragging.

ncjones:
This ensures styles for :first-child and :last-child are applied as expected for remaining list item elements such as when a Bootstrap List Group is used as the sortable list.

### DIFF
--- a/src/html.sortable.js
+++ b/src/html.sortable.js
@@ -12,47 +12,49 @@
 'use strict';
 
 (function ($) {
-  var dragging, placeholders = $();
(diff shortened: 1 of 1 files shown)


ncjones:
I've just realised this is broken for a list with just a single item. I'll try to fix this soon.

lukasoppermann:
Any news here?

ncjones:
Sorry, the project I was using this for was put on hold so I never did the fix. Someone else will have to pick this up if it's desirable.

lukasoppermann:
Closing due to inactivity.

PR #17, opened 2014-03-10
changed dataTransfer.setData based on https://developer.mozilla.org/en-U...

makepanic:
I changed the dataTransfer.setData type to `text/plain`. The old version with `Text` caused my Firefox to go to dummy.com on dragend.
see: https://developer.mozilla.org/en-US/docs/DragDrop/Recommended_Drag_Types


makepanic:
Nevermind http://stackoverflow.com/a/18051912 it wasn't working in IE

PR #18, opened 2014-03-10
added option to define dragImage

makepanic:
This adds the option to define a dataTransfer dragImage for consistency on different browsers.
(I found out that ff on linux displays the dragged element, chromium displays nothing, chrome on win8 shows the dragged element)

see https://developer.mozilla.org/en-US/docs/DragDrop/Drag_Operations#dragfeedback


voidberg:
Merged, thanks!

PR #20, opened 2014-04-24
Call sortable() on the jQuery object instead of the DOM element.

eheikes:
Trying to use an ngModel attribute with the htmlSortable directive gives a JS error.


voidberg:
Thanks!

PR #22, opened 2014-05-06
change dt.setData 'dummy' text for empty string

batista:
Firefox is redirecting to dummy.com because of this text entry. Related to #17.


voidberg:
Thanks!

PR #26, opened 2014-05-31
Add support for jQuery.noConflict version

egorkhmelev:
Btw, angular returns already jQuery powered element to directive, no need to wrap it again


voidberg:
Thanks!

PR #29, opened 2014-06-20
Use jQuery explicit instead of $ as first argument of the wrapper function

ties-v:
By calling the wrapper function explicit with jQuery as first argument,
instead of $, you reduce the chance of library collision. The $ can be
any function of course.


voidberg:
Thanks!

PR #33, opened 2014-07-12
Fix incorrect outerHeight() value when using display: flex

omghax:
This fixes an issue with `forcePlaceholderSize`, where `draggingHeight` was being set after the `dragging` element had been hidden. Doing so causes problems when using `display: flex`, since jQuery's outerHeight function appears to calculate the height as if the element were `display: block`.

Rather than reading outerHeight on every `dragover`, I read it once on `dragstart` and clear it on `dragend`. This fixes my use case, but could cause problems if for some reason the draggable changes height during dragging (not sure when, if ever, this would happen in practice).

I've run into this issue with Chrome 36, but haven't tried it in other browsers.

### DIFF
--- a/src/html.sortable.js
+++ b/src/html.sortable.js
@@ -12,7 +12,7 @@
 'use strict';
 
 (function ($) {
-  var dragging, placeholders = $();
+  var dragging, draggingHeight, placeholders = $();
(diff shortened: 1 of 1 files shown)


voidberg:
Thanks!

PR #35, opened 2014-08-05
Bug when using ng-repeat

sfarthin:
When using something like ng-repeat to iterate through a dataset, connected lists fail. The directive will always incorrectly identify the model "set" to be the same model.

This is because the nested scope changes with ng-repeat and the directive always refers to the root scope of the directive. By retrieving the scope from the starting point element and ending point element of the drag we can determine if this is the same list or a connected list.


voidberg:
Thanks!

PR #37, opened 2014-08-17
Fix placeholder for use with xhtml

coreyfarrell:
The placeholder is currently created as a non-closing tag.  This causes a failure when used in an XHTML page.

Unfortunately I cannot get grunt working, so I haven't updated the files under dist.

### DIFF
--- a/src/html.sortable.js
+++ b/src/html.sortable.js
@@ -48,7 +48,7 @@
 
       var isHandle, index, items = $(this).children(options.items);
       var startParent, newParent;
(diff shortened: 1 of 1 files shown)


voidberg:
Thanks!

molily:
`$('<foo/>')` should have the same result as `$('<foo>')`. This is a question of jQuery parsing the input. It’s not necessary to use `$('<foo/>')` on an XHTML page. jQuery parses both to an element, invoking `document.createElement('foo')`.

http://api.jquery.com/jQuery/#creating-new-elements

@coreyfarrell Did you have actual problems on an XHTML page that uses an X(HT)ML MIME type?

PR #41, opened 2014-09-03
change ordering in angular directive to move element in the array

AdamNowotny:
The current behaviour when dragging elements is to exchange $start element with $end element.
Not sure if that was intended, but what I'd expect is that the element is moved to the new position and the others are shifted.
Example for array [1,2,3,4].
Currently when moving 4 to first position we end up with [4,2,3,1] instead of [4,1,2,3].
It can be observed using angular-single.html example page.

Solution with slice method taken from second answer on [StackExchange](http://stackoverflow.com/questions/5306680/move-an-array-element-from-one-array-position-to-another)


lkstz:
:+1: I can confirm this behaviour, the solution works fine for me. (AngularJS v1.2.22, jQuery v1.11.1)

fiznool:
+1

JeffreyRodriguez:
I was expecting the same behavior when I installed this module and was surprised by the swapping. I reviewed and tested with angular 1.2.18; code looks good and works as described.

:+1:

In the meantime, here's the bower line for @AdamNowotny's branch:

```
"html.sortable": "AdamNowotny/html5sortable#12e23c42a638c670f2c8c19c8912a8fd759b53cb"
```

voidberg:
You're right, not sure what I was thinking when swapping them. :) Thanks!

PR #59, opened 2014-11-28
Move "use strict" inside closure

dstelljes:
This is a really picky fix, but it keeps this plugin from messing stuff up when it's concatenated with other files that don't work under strict mode.


hjellek:
It would be better to remove your style changes and only include the moving of "use strict" since that is what your issue is about.

Other than that, +1, need this fix myself.

dstelljes:
You're right, I should have titled this differently. The other changes are just two inconsequential things that the linter caught. If there's any activity on the repo and it ends up making any difference, this can be closed and #79 merged.

lukasoppermann:
Hey @dstelljes, voidberg got me access to the repo, so we will see some action in the repo now. :)

Fixed in 7fb6b29aa5b5236cbc87e6cbef5882a9573a1d38

dstelljes:
Awesome, thanks!

lukasoppermann:
Sure, thanks for bringing it up. :+1:

PR #62, opened 2014-12-08
Create 'clone' option to prevent object from being removed

tlshaheen:
Setting the option 'clone' to true will **duplicate** (or clone) the object to the new location (in a different sortable list), rather than **move** then object to the new location. Defaults to false of course.

Simply set `$('original-list').sortable({ clone: true });` on the original list, and `$('new-list').sortable({ clone: false, connectWith: 'original-list' });` on the target list.

This is useful when doing something like a WYSIWYG editor (https://github.com/farhadi/html5sortable/issues/100).

You'll see I'm carrying `originaldragging` and `dragging` both throughout the drag process. Some times you want to reference the actual object that isn't being moved, some times you want to reference the new clone/copy.


tlshaheen:
The original commit only worked for specific cases, and not very many of them. The new commit ef428ff9f3235758405d635da349abdfd9f43f04 works for all my test cases. New items cannot be added to `clone: true` containers, and `clone: true` elements will not replicate themselves when being re-arranged in a `clone: false` container.

lukasoppermann:
Hey, @tlshaheen can you please give me an example for when you would actually use something like this. I would like to understand the problem you are trying to solve, to come up with a good solution.

tlshaheen:
@lukasoppermann Anywhere you would drag and drop components into a WYSIWYG editor. You could be creating a webpage, or creating an email like at https://templates.campaignmonitor.com/canvas/design/01-mason#/firstrun

At the above link, html5sortable could be used to drag the image or text button into the email.

lukasoppermann:
@tlshaheen, okay, sounds valid.

Could you do me a favour and do the following.
- pull the current master and add changes to this version
- add some tests
- only change the files needed (src/html.sortable.js & test files) (it has bend change so that you will be added to the contributors once your PR is merged and in a release automatically.)
- squash to only 1 commit
- add some info the the readme so users know about and understand this new behavior

Thanks

lukasoppermann:
Closing due to inactivity.

PR #64, opened 2014-12-19
Fix height for the placeholder

jacobsvante:
Using outerHeight can cause elements to become too big, pushing down floated elements. I've noticed this happens when I use `box-sizing: border-box` but might apply otherwise also.

### DIFF
--- a/src/html.sortable.js
+++ b/src/html.sortable.js
@@ -81,7 +81,7 @@
         }
 
         index = (dragging = $(this)).addClass('sortable-dragging').attr('aria-grabbed', 'true').index();
-        draggingHeight = dragging.outerHeight();
+        draggingHeight = dragging.height();
         startParent = $(this).parent();
         dragging.parent().triggerHandler('sortstart', {item: dragging, startparent: startParent});
       }).on('dragend.h5s',function () {
@@ -109,7 +109,7 @@
           e.preventDefault();
           e.originalEvent.dataTransfer.dropEffect = 'move';
           if (items.is(this)) {
-            var thisHeight = $(this).outerHeight();
+            var thisHeight = $(this).height();
             if (options.forcePlaceholderSize) {
               placeholder.height(draggingHeight);
             }


lukasoppermann:
I think this is not a good choice, since it does not include padding and border. (http://api.jquery.com/height/)

You would either need to add those or otherwise use outerHeight and subtract a margin. 

Do you have an example of the problem you are describing?

jacobsvante:
I know it's not including padding and boarder. I use a custom placeholder that inherits the tag and classnames that the items have. I'll take a further look at this to see if I can find a more permanent solution.

lukasoppermann:
I actually saw the problem you describe with the grid.

The issue is though, that it depends on the whole setup. If the user uses a border you need to not include this, if he uses a no border, it would need to be included. Same with padding.

So neither using outerHeight, nor height can solve this problem. Maybe we can check whatever the user sets for the placeholder and calculate this? Like include border? This would need to be documented.
I like the first idea better, but maybe its to much of an overhead.

jacobsvante:
I think that could be solved by forcing [box-sizing: border-box](https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing)

lukasoppermann:
True we could set it via js but still the user could accidentally overwrite it. However the support is quite good http://caniuse.com/#feat=css3-boxsizing.

I am just not so sure if a js only solution would be better. But maybe this is really the way to go.

A problem though is if we set it to border box and the user sets the padding on the placeholder to 0. Than the sizes is to small.

jacobsvante:
Yeah support is really good for it. Considering we set the height with JS today adding `box-sizing` is quite natural IMO.

lukasoppermann:
I think there is a different. If you use `forcePlaceholderSize`you expect the height and maybe width to be set. However you do not necessarily expect other attributes to be set as well.

Maybe it is all just a case of good documentation though, and your initial idea was the best, using height but adding to the documentation, that the placeholder will only get the height of the elements excluding borders, padding, etc.

However did you check if this leads to problems because the draggingHeight is used for calculation of the deadZone. So maybe changing it to height causes problems there.

jacobsvante:
You're probably right. Haven't had time to look it up further. It would be great if some more of you guys could test this and see if you notice any regressions.

> However did you check if this leads to problems because the draggingHeight is used for calculation of the deadZone. So maybe changing it to height causes problems there.

I'm not super-familiar with the code and it's kind of hard to follow when there's no comments. What problems could we get with the dead zone? I use this fix in production and it's been working out great so far :blush:

lukasoppermann:
I totally agree with your comment on the "missing" comments. I would like to make the code more dev friendly, by not using chaining so extensively and adding some comments. What do you @voidberg think about that?

The dead-zone code is supposed to stop flickering, which basically happens when the placeholder is so small, that your cursor could be above or below it. I did however not look into this very deeply.

I will test this fix and maybe add some comments on what you could add to the docs for this change (so the changed docs can be with this commit as well).

jacobsvante:
I can agree that the chaining in the current code makes it kinda hard to read. Fixing that plus some more comments would go a long way. Maybe breaking out into more functions and documenting them in JSDoc format.

Hmm I haven't noticed that about the flickering. Thanks in advance for testing the code.

lukasoppermann:
I tested it and did not find any problems.

I suggest you add some more info to the docs where it talks about `forcePlaceholderSize`.

Like:

This will set the size of the placeholder to the height of the dragged element. This does not include any padding, border or margin. You will need to define your css so that those attributes are added to your placeholder.

You can style the placeholder using the `.sortable-placeholder` class.

@jmagnusson feel free to change this completely, I just wanted to show what I think would be helpful in the docs.

PR #67, opened 2014-12-19
update example.html to include demo for nested list

lukasoppermann:
also separate styles classes & js functionality, this makes it more easy to add new items without having to deal with styling issues.

### DIFF
--- a/examples/examples.html
+++ b/examples/examples.html
@@ -7,15 +7,11 @@
 		header, section {
 			display: block;
 		}
-		#features {
-			margin: auto;
-			width: 460px;
-			font-size: 0.9em;
-		}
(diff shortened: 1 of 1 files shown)


examples/examples.html
```diff
@@ -53,16 +51,12 @@
 			background: #FEE25F;
 		}
 		#connected {
-			width: 440px;
+			width: 620px;
```
jacobsvante: Would be nice if you could change this to `max-width` to make stuff responsive.

lukasoppermann: Done.

jacobsvante: :+1:

lukasoppermann:
Added another test-case for a disabled list.

Issue #66, opened 2014-12-19
Nested lists broken since 0.1.7


lukasoppermann:
I noticed nested lists do not work, even though they do in this example #55. The bug was introduced in 0.1.7 f7e1b1454356de816e6aef2cf72e3f9a7a8c9c5c

The bug which 0.1.7 fixes is present in chrome & firefox without this fix, so we can not just revert it.

Any ideas why @splagemann?

PR #68, opened 2014-12-20
fixing nested list bug

lukasoppermann:
Fix for #66.


jacobsvante:
Works great for me :+1:

PR #70, opened 2014-12-22
Example with nested lists

jacobsvante:
Nested lists have been broken for a while as is noted in #66 and
is fixed by @lukasoppermann in #68.
Added this example so that we can test for regressions more easily in the future. This commit also makes the example page responsive.

### DIFF
--- a/examples/examples.html
+++ b/examples/examples.html
@@ -9,13 +9,13 @@
 		}
 		#features {
 			margin: auto;
-			width: 460px;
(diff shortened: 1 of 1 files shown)


lukasoppermann:
Hey,
I did already add a nested list in #67.

jacobsvante:
Ah sorry, I totally missed that.

lukasoppermann:
No worries.

PR #77, opened 2014-12-27
fix ghost for handles & adding advanced functionality

### LINKED ISSUE #75: Ghost is broken when using handle.
When using a handle only the handle item is used as a `ghost` while it should be the whole item.

You can test it here http://htmlpreview.github.io/?https://github.com/voidberg/html5sortable/blob/master/examples/examples.html

lukasoppermann:
This PR fixes #75 by creating a `ghost` from the dragged item.

Additionally it makes it possible to customise the looks of the dragged item when using the `ghost` option form `dragImage`

Those options are added to the docs.

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -56,12 +56,13 @@
 		    </div>
 				<div class="col col-6">
 					<ul class="ml4 js-sortable sortable list flex flex-column list-reset">
-						<li class="p1 mb1 navy bg-yellow">Item 1</li>
-						<li class="p1 mb1 navy bg-yellow">Item 2</li>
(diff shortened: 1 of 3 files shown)


lukasoppermann:
This needs to be tested:
- [x] IE 9
- [x] IE 10
- [x] IE 11
- [x] FF
- [x] Chrome
- [x] Safari

**IE Fix:**

IE 10 - 11 can use a fix which is like this:

``` javascript
if (typeof document.createElement("span").dragDrop === "function") {
   handles.on('mousedown', function(e){
      this.dragDrop();
   });
}
```

It does however not work for IE 9

PR #82, opened 2015-02-01
Add sortchange event

kausters:
Added a `sortchange` event that activates when moving the dragged item, once it's moved into a position without being yet released. This could be used, for example, to dynaimcally update the placeholder based on its position or the dragged item. A sample use would be setting the placeholder's size to match the dragged item.

This PR doesn't include changes to the Angular version though.


lukasoppermann:
Hmm, interesting idea, but couldn't this be done without `mouse events` but by adding it to the `dragenter` or something like this? We are updating the placeholder anyway, so we might fire the event than.

I do not want to add to many additional jquery event calls and I think it is actually not needed.

I would appreciate if you could refactor this with the following aspects:
- add this to `dragover.h5s dragenter.h5s` instead of using mouse events
- don't bump the version
- don't change the Authors file (both will be done automatically)
- don't change any files in the dist directory (so basically only readme & the file in src)
- rebase to one commit with a descriptive name

**Please update to the latest version via git pull and add your changes there otherwise merging will be hard.**

lukasoppermann:
Hey @kausters will you be editing this and update your PR? Would be awesome.

lukasoppermann:
Hey @kausters can you please tell me if you are planning on refactoring this PR so it can be merge?

lukasoppermann:
Closing due to inactivity.

PR #83, opened 2015-02-01
Horizontal deadzone

kausters:
Stops flickering when dragging horizontally (float/inline-block) placed items.


lukasoppermann:
Hey @kausters do you have an example when this happens (how can it be reproduced)?

voidberg:
This is actually valid. I don't have a handy example but I ran into it while reordering a list of images rendered by [rowgrid.js](http://brunjo.github.io/rowGrid.js/). Initially, all items have the same height but then rowgrid adjusts it so they fit nicely on each row, causing each row of items to have variable height and sortable to act odd. Implementing this fixed it.

@kausters are you able to refactor the PR? I can do it if you don't want anymore.

lukasoppermann:
Hey @kausters, could you do me a favour and update your PR like this:
- merge it with the current version
- only change and commit `src/html.sortable.js`

Please don't bump the version, we will do this once we release a new version. Thanks for contributing.

lukasoppermann:
Hey @kausters any chance you can rebase this?

PR #87, opened 2015-03-13
Hover state class rather than relying on :hover

brandonferens:
Added in a hover state class if mouse is over draggable element. Using :hover causes issues if dragged element is over the next element. Using js to apply or remove a hover class solves this issue


lukasoppermann:
What happens though if we have nested lists? Also what happens on mobile?

brandonferens:
I'm not sure about nested lists as I don't have a need for that, however, nothing happens on mobile. I guess I need to look at that code again as I need it to function on mobile as well...

lukasoppermann:
That would be good. Could you do me a favour and and do a pull before you change anything? 
Do you have a guess on a performance influence? Should it maybe be optional via config, default to off? Since it is possible not something many people use.

brandonferens:
I will definitely do a pull before I mess with the code again. The original code based didn't really have any sort of hover functionality so I can see how people might not use that, but for UX purposes, we felt it necessary for our use. Perhaps an optional config enabled via either js or a data attribute may be beneficial.

lukasoppermann:
Hey, I do see a point in UX, I am just worried it could add a negative performance bump for people who never use it. I though something like this:

In the config array add this:

``` javascript
    options = $.extend({
      connectWith: false,
      placeholder: null,
      dragImage: null,
      hoverClass: false
    }, options);
```

Now do a check in the beginning, if it is false, don't add it, if it is anything else, use this as a class.

So the user can do `hoverClass: 'drag-hover'` and this will be added on hover.

brandonferens:
I think that is perfect.

lukasoppermann:
Hey @brandonferens any news on this PR?

brandonferens:
Sorry @lukasoppermann. I've been heads down with work and haven't been able to get to it yet.

ghost:
I dropped this bit of code in to fix an issue where the `:hover` state was getting stuck. See the animation below for an illustration of the problem.

lukasoppermann:
Hey @brandonferens any news on this? Did you check out the mentioned error? @uranther did you find a solution for it?

ghost:
@lukasoppermann The solution for my problem is the same as the 5 lines of code in this pull request. Looks like we just need to fix the merge conflict so it can be merged.

lukasoppermann:
Ahh, so you die not have The error AFTER The fix?

ghost:
Correct. I took out my CSS that used the `:hover` selector and changed it to the `.sortable-over` selector, and that fixed the problem shown in the GIF above.

jaruba:
this also fixed my hovering issues

jaruba:
actually.. it didn't.. this seems to be a chromium issue, the proposed solution only fixes hover for the list items after dropping an item but everything else on the page still has this bug until the user clicks somewhere in the page.

lukasoppermann:
Hey @brandonferens I would like to merge this, you you just very quickly do the following:
- update your branch to the latest version
- only change `src/html.sortable.js`
- add a section to the docs?

Otherwise I will just create a new PR and add your changes, but I would rather just merge a cleaned PR from you, since it was your idea. Thanks.

brandonferens:
Sure, I can do that. I will do it as soon as I can!

lukasoppermann:
Hey @brandonferens can you get onto this? (It should not take you so long) so I can merge this, as it seems to fix some bugs for people.

brandonferens:
Do you want me to run gulp as well?

lukasoppermann:
nope, just the tests so only `npm test` but please make sure to work with the current code base, thanks. :+1:

brandonferens:
There you go @lukasoppermann. Sorry it took so long. I could give you all the reasons in the world, but I just plain failed to get on this. Hope this new code helps you out!

lukasoppermann:
Thanks @brandonferens :+1: I will merge it later today.

lukasoppermann:
Hey @brandonferens did you by any chance change the indentation? ;) could you please undo this and just change the effected lines? Otherwise it is pretty hard to merge, thanks.

brandonferens:
I didn't. At least I don't think so. You have 2 spaces as indent, correct? Which file(s) are giving you issues?

lukasoppermann:
Well, let me check once I am at my Mac maybe it's just an issue when
viewing it by phone, but it looks like nearly all lines are changed.

lukasoppermann:
Hey, the problem is, that the file is actually called `html.sortable.src.js` and you added the file called `html.sortable.js` could you please update your PR? Thanks

brandonferens:
I must have done that awhile ago and just built the file for my own use. I have removed it and updated the src file.

PR #96, opened 2015-04-22
Added UMD pattern (WIP)

lukasoppermann:
It seems to be working fine in the browser, but I did not test everything yet.
- [x] browser through globals
- [x] test amd
- [x] test module.exports (define) 
- [x] test with angular
- [ ] ~~test with es6~~ 

It seems a little problematic assigning it to the $.fn, maybe use a function instead?

``` javascript
htmlsortable('.class', {...});
```

**No ES6 for now**
Sadly I did not find any way of including an es6 module live into a website. Without this, it creates a development overhead without any gain (since you would need to compile the file before checking if it works).

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -34,7 +34,11 @@
     "tests"
   ],
   "devDependencies": {
-    "jQuery": "~2.1.3",
-    "angular": "~1.3.15"
+    "angular": "~1.3.15",
+    "requirejs": "~2.1.17",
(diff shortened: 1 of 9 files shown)


topguns:
What the codes to get feed backs and where to put them at im so lost

lukasoppermann:
Sorry, I don't really get you. What are you asking?

PR #100, opened 2015-04-28
Removed the need for a globally defined jQuery variable

jacobsvante:
Also added some tests for this to make sure we have no regressions in the future.

My primary motivation for this PR is for html5sortable to work with Browserify without shimming. With the latest code at the time of writing I get `Uncaught ReferenceError: $ is not defined`. I wonder if this code really had been tested in a node / common JS environment.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -13,7 +13,7 @@ var jshint = require('gulp-jshint');
 var jscs = require('gulp-jscs');
 var uglify = require('gulp-uglify');
 var sourcemaps = require('gulp-sourcemaps');
-var umd = require('gulp-umd');
+var mochify = require('mochify');
 /* ---------- */
 /* error handling */
 var reportError = function(error) {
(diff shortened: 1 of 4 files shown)


lukasoppermann:
Hey, this plugin was only tested with the require1k plugin.

I would like to merge this, but I have a lot of questions and requests first. ;)
1. is there any reason for removing the umd gulp task (please do not, if there is not definite need, as it helps simplifying the src)
2. Why did you link the unminified dist in the package.json is that the common thing to do?
3. Why do you return jquery in the end?
4. If I clone your repo, run `npm i`and `gulp test` 2 tests fail. The first is easy to fix, but the second is not for me. Any help?
- [ ] please do not remove the linting from the test task
- [ ] please remove all indentation changes, this just make it impossible to see where you actually change something

I will probably have more questions once you answered those and changed the 2 requested things. I do really appreciate you adding tests (this is on my to do list anyway). If you feel inclined to add more tests, feel free. :)

lukasoppermann:
Actually, updated the umd gulp-function a bit and now your test passes just fine. I did already add phantomjs and mochify as well as the test/ dir so you can just get the current version of the repo and add your test to the test folder. Do an `npm test` and it should pass.

Please update this PR to only include the test. Thanks.

jacobsvante:
Does the gulp-umd actually provide any value? Isn't it better to be in control of the output yourself? At least that's how I feel about code in general. Especially when it's this simple. IMO jQuery doesn't have to and shouldn't be a dependency at all to this project, but that's a whole other case.

lukasoppermann:
Hey @jmagnusson it does provide a value, less confusion for developers who are not used to umd & removing a potential source of errors. 

Are you talking about jquery at the status quo? Because we are using it all through the project, so it needs to be a dependency. Or are you saying it should be removed from the plugin all together?

jacobsvante:
Isn't it more confusing that the code is dynamically altered in production? If you're a JS-programmer these days you need to know about UMD. It's not exactly magic! :-)

I mean that - just like `gulp-umd` - `jquery` doesn't provide any real value to the project and that it would be nice to get rid of it all together. Keeping stuff simple is nice ;-)

lukasoppermann:
While I do disagree on both I would like to remove the need for jquery as well.

But it does solve many problems for you nicely. I am planning on removing at least the need for jquery in the long run, but this is a tough task and will need much preparation.

As for gulp-umd I guess this is a question of taste, like many other topics. I prefer to keep environmental details out of the source, but I understand that you might favour something else. :)

In any case I would appreciate a PR with your browserify test.

jacobsvante:
You don't have to agree on everything! The meaningfulness of jQuery these days [can be questioned though](http://youmightnotneedjquery.com/).

PR #107, opened 2015-04-29
Fix and test browserify compatibility

jacobsvante:
As per @lukasoppermann's request in #100.

Probably useful for other node-using bundlers as well.

Note that I had to turn off `maxstatements` because it made the linting process fail at that test. At least with a fresh `node_modules` directory.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -15,7 +15,7 @@
   "regexp": true,
   "maxdepth": 3,
   "maxparams": 2,
-  "maxstatements": 25,
+  "maxstatements": 0,
   "maxcomplexity": 10,
   "undef": true,
   "unused": true,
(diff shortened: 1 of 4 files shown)


lukasoppermann:
Hey, could you do me a favor and turn the max-statement on again? The failing has nothing to do with your code, so I can pull it anyway. 

I added a jshint comment to the src for now, so the error does not pop up, but it is something that should be fixed, so I do not want to change the `jshint` file

jacobsvante:
There ya go.
Also noted that you don't test against the latest node version in travis so added that (specifying `node` in the config means "use the latest version").

lukasoppermann:
Hey, thanks, yes I did see you updating this, good catch, thanks.

PR #118, opened 2015-05-08
move all tests to mocha + jsdom

lukasoppermann:
All tests are moved to mocha with jsdom. This way they all run in one environment, which gives us the benefit of a more realistic coverage and also less dependencies.


lukasoppermann:
@jmagnusson since you seem to know more than I do about browserify: all it requires is a correctly setup commonJS module, right? If so, the way I migrated your tests to use jsdom without browserify should still suffice, correct? If so, I would merge this. Thanks for your input.

jacobsvante:
Works just fine for me in browserify :+1:

lukasoppermann:
Okay, thanks. :+1:

PR #133, opened 2015-05-20
Fire an event when the user lets go 

polyfloyd:
The existing `sortupdate` event is only fired when items are moved around.

This is a problem when one wants to switch some UI components on and off when the user is dragging items around: if the user starts dragging an item and drops it in the starting position, the program waiting for the `sortupdate` event will never receive it, leaving the UI in a broken state.

I added the `sortstop` event to address this problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,22 @@ $('.sortable').sortable().bind('sortstart', function(e, ui) {
 });
 ```
 
+### sortstop
+Use the `sortstop` event if you want to do something when sorting stops:
+
+``` javascript
+$('.sortable').sortable().bind('sortstop', function(e, ui) {
+    /*
+
+    This event is triggered when the user stops sorting. The DOM position may have changed.
+
+    ui.item contains the element that was dragged.
+    ui.startparent contains the element that the dragged item came from.
+
+    */
+});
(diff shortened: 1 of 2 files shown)


lukasoppermann:
I see, this does make sense, but maybe we should include the oldindex as well?

I was thinking if we could change sortupdate to include a flag `moved` or something, but it is probably better to have a separate event.

Can you please to the following so I can pull it:
1. undo the version bump
2. undo the change to Authors
(Both will be done automatically once the PR is added to the new version)
3. remove the changes from `dist/html.sortable.js``

Please make sure your PR has only one commit if you have more, rebase to one commit with a sensible name. `Add the sortstop event to sortables` is good.

polyfloyd:
All done.

Also, the readme did not mention that the AUTHORS and version number are updated automatically. Shouldn't it be updated?

lukasoppermann:
Thanks, will merge it later on. 

I did change the readme yesterday to state that you do not need to add yourself to the Authors file. :)
I will add the part with the version number as well though, good point.

polyfloyd:
@lukasoppermann When will you publish the new version containing these changes?

PR #143, opened 2015-06-12
Allow the usage of controllerAs. Closes #86.

### LINKED ISSUE #86: Angular directive doesn't support "controller as" syntax
Working with traditional use with $scope:
http://codepen.io/ronnieoverby/pen/vEjroy

Not working with controller as syntax:
http://codepen.io/ronnieoverby/pen/RNyBNp

This is unfortunate for me because the way your executing the callback is forcing me to deviate from the normal architecture of my app and use $scope.

[An approach that would probably solve the issue](http://stackoverflow.com/a/19960066/64334) (but not be backward compatible) is to parse the callback expression and force the developer to be explicit about arguments.

voidberg:


voidberg:
@lukasoppermann Want to double check?

lukasoppermann:
@voidberg sorry, I never worked with angular, so me checking will not help you. :) If you would add some tests for the angular part I would be very happy though, as I did not add any, due to not knowing angular.

PR #152, opened 2015-06-23
hide dragging once and append to end

shafiry:
There are two parts to this change:
The first is that dragging.hide() happened over and over again during sorting, which isn't an issue, but also is unnecessary. 
The second is that the hidden list item breaks certain grids that work by checking the number of `<li>`s, because it thinks the display none item and the placeholder take two separate spots. I fixed this by moving the dragging item to the end, where it can not affect the grid. Here is an example using foundations grid (drag 1, and see that 3 moves down to the next row):

http://jsfiddle.net/gRtrX/228/

### DIFF
--- a/src/html.sortable.src.js
+++ b/src/html.sortable.src.js
@@ -368,8 +368,13 @@ var sortable = function(selector, options) {
         }
 
         if(dragging.is(':visible')){
+<<<<<<< Updated upstream
           dragging.appendTo($(dragging).parent());
           dragging.hide();
+=======
+          dragging.hide();
+          dragging.appendTo($(dragging).parent());
+>>>>>>> Stashed changes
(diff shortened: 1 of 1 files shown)


lukasoppermann:
Hey @shafiry,

thanks for your contribution. I will try to check it out and test it as soon as I can, it sounds pretty good. :+1:

shafiry:
Glad to hear!

shafiry:
Just realized that line 371 and 372 should be flipped (hide before appending to end). How should I go about updating the pr?

lukasoppermann:
Hey @shafiry just update it in your branch and rebase to one commit. It will be automatically be updated here.

shafiry:
continued in #155

PR #155, opened 2015-06-26
hide dragging once and append to end

shafiry:
This continues pull request #152


shafiry:
Fixed an issue where "cancelling" a drop by ending outside of the list failed.

lukasoppermann:
Hey, please take care of the following:
1. do not commit dist files 
2. rebase to only commit with a good name
3. If at all possible, please add test for it

thanks.

shafiry:
continued at #158

PR #158, opened 2015-06-30
hide dragging only once and append to end of list for grid fix

shafiry:
Continuation of #155 
This change was made because of grid views when using foundation (and potentially other frameworks) that rely on number of list items to determine placement on a list, the placeholder and hidden dragging item both take space so it throws off the grid. See jsfiddle:

http://jsfiddle.net/gRtrX/228/

The new global variables are introduced because just appending to the end of the list at the beginning, it must then be returned to it's original position in the case of cancelling a sort by either placing outside of placeholder space or pressing 'esc'.

### DIFF
--- a/src/html.sortable.src.js
+++ b/src/html.sortable.src.js
@@ -83,6 +83,13 @@ var _makeGhost = function($draggedItem) {
   };
 };
 /**
+ * _clearPlaceholders sets the placeholders global variable to 0 to
+ * fix is('visible') issue with jsdom
+ */
+var _clearPlaceholders = function() {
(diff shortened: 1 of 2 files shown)


lukasoppermann:
Hey @shafiry any chance you could rebase and add some tests?

lukasoppermann:
Hey @shafiry can you try to explain what you are doing with the 2nd test? And What kind of bug you are fixing with this? I am trying to port it to the new version, but I have trouble following the changes.

lukasoppermann:
Closing due to inactivity.

Issue #161, opened 2015-07-11
document is not defined


jaruba:
I installed `html5sortable` through npm:

```
npm install html5sortable
```

in my module I set:

```
require('html5sortable');
```

the first time I use `.sortable()` I get:

```
Uncaught ReferenceError: document is not defined
```

I know for a fact that in NW.js modules the `document` object is located at `window.document`.
Adding:

```
typeof window.document.createElement('span').dragDrop === 'function'
```

to this line: https://github.com/voidberg/html5sortable/blob/master/dist/html.sortable.js#L177
fixes this issue.

lukasoppermann:
Should be fixed by #165

PR #165, opened 2015-07-24
fix for #161

lukasoppermann:
Using window.document when document is not available. Also small refactoring to fit into 80 char styleguide rule.


lukasoppermann:
Currently it fails due to coveralls being in `read only`.

PR #192, opened 2015-12-18
Add support of binding ngModel via data-ng-model

jamesjieye:
Currently, if you add `data-` to before `ng-model`, the sorting result is not updated to the model. It would be nice to support `ngModel` binding via `data-ng-model`.

### DIFF
--- a/src/html.sortable.angular.js
+++ b/src/html.sortable.angular.js
@@ -45,8 +45,8 @@
               }, true);
 
               element.sortable().bind('sortupdate', function(e, data) {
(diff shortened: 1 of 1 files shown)


lukasoppermann:
Hey @jamesjieye, thanks for your contribution so far.

 I do not know angular, is what you are prosing a common angular pattern or a personal preference? If it is common, could you please link docs or an article where I can read about it?

lukasoppermann:
Okay, seems to be common, as it is also mentioned in the docs. https://docs.angularjs.org/guide/directive

Issue #181, opened 2015-09-30
Handle doesn't count when dragging begins from a link


akondo06:
Hi,
I got to use the plugin in a project and was wondering why this was happening.
Check:
https://jsfiddle.net/5cg42sfo/3/

When you **drag the link**.. the **dragstart** is triggered. Not sure if it's a plugin issue.
Tested in **Firefox 40.0.3** and **Chrome 45.0.2454.101 m**

Looking forward for the fix!

akondo06:
Made it work by setting attribute **draggable** to **false** on the link element. 
Still .. should be handled in a better way.

Mushr0000m:
Hi, Same issue with img tag. I have set the handle param, but for some reason, if there is an img tag inside my sortable element list (in my case it's for a gallery) then the img tag still act as a draggable element. So need to add draggable="false" on the img. Then the handles works.

lukasoppermann:
This should be fixed, if it persists, please reopen.

Issue #166, opened 2015-07-29
Restrict dragging to handle


bennyschudel:
When there is another element with > draggable="true" < within the element, restricting dragging to the handle does not work anymore. I guess it needs a check on dragstart if it is the handle element or not.

kshipp:
Doesn't work on mobile.  :(

lukasoppermann:
Are you talking about nested elements. This seems to work fine on my machine. If this issue persists, please reopen with a demo showing the issue, thank you.

Issue #147, opened 2015-06-16
Any anchor tag acts as a handle


shafiry:
I placed a handle specifcally for ".move-button" classes, of which every list element has a button inside with that class. The list elements also have an anchor tag inside, which when grabbed act as handles. Any idea why this would be happening?

lukasoppermann:
Can you please upload an example on jsfiddle?

shafiry:
https://jsfiddle.net/9quqw7fo/22/

the anchor acts as a handle when it probably shouldn't

lukasoppermann:
Actually it seems like the problem is also applying to selected elements, maybe it gets into a "draggable-like" state. To prevent this we probably have to check if the currently dragged element is within the handles array and if not abort https://github.com/voidberg/html5sortable/blob/master/src/html.sortable.src.js#L266

lukasoppermann:
Actually, while it does not work anymore there is an error thrown, when dragging a non-handler element.

https://jsfiddle.net/9quqw7fo/22/

```
(index):736 Uncaught TypeError: Cannot read property 'parentElement' of null
    at HTMLUListElement.onDragOverEnter ((index):736)
```

We need:
- [x] fix for the issue
- [ ] test to validate the fix

lukasoppermann:
Should be fixed, if still persists, please reopen with demo.

PR #202, opened 2016-03-17
Fix non-handles acting as handles.

lizzielizzielizzie:
Elements with draggable="true" initiated sorting, regardless of whether
or not they were the intended handle.


lizzielizzielizzie:
As before, this should resolve issues #181 #166 and #147

Additionally you mentioned commiting any changes I made to the angular file, but it doesn't appear that any change is necessary. In all fairness though, I have literally no experience with angular

lukasoppermann:
Hey @NickWer there should be no change, your are right. I thought there was, because previously you committed the angular file from `dist` but this was only due to it being recompiled with the new main source file.

I will check it out later and merge it in. Thank you for the contribution!

lukasoppermann:
Hey, I checked it out, thanks. I would like to merge the PR, however, I would need to ask you to do the following before I can merge it:
- correctly indent the code https://github.com/voidberg/html5sortable/pull/202/files#diff-f42dd2e6f1fc935a2fd8a4911d6dfb1aR286
- if possible, add the cases described in #181 #166 and #147 to the `index.html` so as to verify they are really solved by this.

Thanks a lot,
Cheers

lukasoppermann:
Ping @NickWer

lizzielizzielizzie:
I addressed the items you mentioned, as well as updated them to meet the requirement that jquery be removed. Unfortunately I had to remove the test cases I had written, as I was unable to simulate a dragstart event on the handles, but added a test case to the index.html

Note that what may seem like (at least it did, to me) an extraneous ondragstart method added to index.html, [is required to actually cause the event to fire](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/draggable).

PR #204, opened 2016-06-16
Without jquery

### LINKED ISSUE #60: VanillaJs
Hey @voidberg,

what do you think about ditching frameworks and just make it work with vanillajs?

Maybe in a way that people who want jQuery for backwards compatibility could plug it in?

nazar-pc:
Large step by step refactoring to eliminate jQuery usage completely.
Fixes #60

Library will need IE10+ (rather than IE9+ previously) because of `Element.classList`, but with polyfill it should work even on IE9. `jsdom` was also updated to include some necessary features not available in older releases.

jQuery support dropped entirely, but to be honest, `sortable($element.get())` is not that difficult :smile:

Major differences:
- instead of jQuery collections native elements, `NodeList`s or arrays of native elements are used/supported
- events data are now in `e.detail`
- events are fired on all elements from group (when using `connectWith`)
- selectors are limited to native

### DIFF
--- a/src/html.sortable.src.js
+++ b/src/html.sortable.src.js
@@ -41,6 +41,10 @@ var _removeData = function(element) {
     delete element.h5s.data;
   }
 };
+/**
+ * Cross-browser shortcut for actual `Element.matches` method,
+ * which has vendor prefix in older browsers
+ */
 var matches;
 switch (true) {
   case 'matches' in window.Element.prototype:
--- a/src/html.sortable.src.js
+++ b/src/html.sortable.src.js
@@ -99,6 +99,13 @@ var _off = function(element, event) {
     delete element.h5s.events[event];
   }
 };
+var _offset = function(element) {
+  var rect = element.getClientRects()[0];
+  return {
+    left: rect.left + window.scrollX,
+    top: rect.top + window.scrollY
+  };
+};
 /*
  * remove event handlers from items
  * @param [jquery Collection] items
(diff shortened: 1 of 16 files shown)


src/html.sortable.src.js
```diff
@@ -192,6 +192,14 @@ var _reloadSortable = function(sortable) {
   // remove event handlers from sortable
   _removeSortableEvents(sortable);
 };
+/**
+ * Get position of the element relatively to its sibling elements
+ * @param {Element} element
+ * @returns {number}
+ */
+var _index = function(element) {
+  return [].indexOf.call(element.parentElement.children, element);
```
lukasoppermann: I think you'd want `Array.prototype.indexOf` here as it does not create an empty array which has to be garbage collected.

nazar-pc: I thought it have negligible impact, so I decided to use which is shorter, but if you wish so - no problem

lukasoppermann:
Hey, just letting you know. I appreciate the huge amount of work. But since you changed so many files I will need some time to look through it, until I can merge it with confidence.

nazar-pc:
I understand the process, not a problem:)
If you find some decisions unclear or prefer something to be done other way - I'm always open for comments.

src/html.sortable.src.js
```diff
@@ -14,30 +14,170 @@
...
+/**
+ * Remove data from element
+ * @param {Element} element
+ */
+var _removeData = function(element) {
+  if (element.h5s) {
+    delete element.h5s.data;
+  }
+};
+var matches;
```
lukasoppermann, line 48: Please add a docblock or comment here, so it is immediately clear why this switch statement is needed. (Only for the `_filter` below?)

src/html.sortable.src.js
```diff
@@ -14,30 +14,170 @@
...
+    }
+    return;
+  }
+  if (event.indexOf(' ') !== -1) {
+    for (i = 0, events = event.split(' '); i < events.length; ++i) {
+      _off(element, events[i]);
+    }
+    return;
+  }
+  if (element.h5s && element.h5s.events && element.h5s.events[event]) {
```
lukasoppermann, line 109: Wouldn't it be enough to just have the last `if`-statement in the `_off` function? This would mean you would need to call `_off` on every element, for every event, but it would also make the code easier to skim. 

I feel the function mimics jquery behavior but we should rather keep it a little more simple, event if we have to repeat function calls.

src/html.sortable.src.js
nazar-pc, line 48: Yes, it is only used in `_filter()`, but since:
1. it is better to compute once
2. Line in `_filter()` is pretty complex already

I decided to give it separate variable. Comment added in last commit.

src/html.sortable.src.js
nazar-pc, line 109: I think we can drop support for space-separated events list, but I'd like to keep first `if` so that it would be possible to remove event handler from array of elements (which will become quite lengthy otherwise). What you think?

nazar-pc, line 109: Look at last commit if it is better than before.

lukasoppermann, line 109: Not completely convinced we need this, but lets just remove the **space-separated events** for now. We can still decide on the other part once this is merged.

lukasoppermann:
Hey @nazar-pc it looks very good. If only found one minor problem, which is actually in the demo. You need to update the `sortupdate` demo to use e.originalEvent.detail, like so:

```
console.log('Parent old: ');
console.log(e.originalEvent.detail.startparent);
console.log('Parent new: ');
console.log(e.originalEvent.detail.endparent);
console.log('Index: '+e.originalEvent.detail.oldindex+' -> '+e.originalEvent.detail.index);
console.log('elementIndex: '+e.originalEvent.detail.oldElementIndex+' -> '+e.originalEvent.detail.elementIndex);
```

Lines 268+ in `index.html`.

nazar-pc:
Fixed in last commit

lukasoppermann:
Thank you very much, I will do the Pull tomorrow.

nazar-pc:
I didn't change version, but this is definitely a BC break, so major version should increase (or minor since it is still under 1.0)

PR #208, opened 2016-07-12
[Issue] Unable to drag & drop items to a shorter list without having to scroll

marrouchi:
Hi,

There is a problem when trying to move a draggable item from the bottom of a long list (A) to another shorter list (B), the placeholder stuck in A list even if the dragged item is over the B list. It works when the B list is empty, but if it contains at least one item then you need to scroll up and put the mouse pointer over one of the items.

https://jsfiddle.net/L247db7h/

Regards,

### DIFF
--- a/src/html.sortable.src.js
+++ b/src/html.sortable.src.js
@@ -649,7 +649,7 @@ var sortable = function(sortableElements, options) {
           .forEach(_detach);
       } else {
         if (placeholders.indexOf(this) === -1 && 
-          (!_filter(this.children, options.items).length || e.toElement.className != options.placeholderClass)) {
+          (!_filter(this.children, options.items).length || e.toElement.className !== options.placeholderClass)) {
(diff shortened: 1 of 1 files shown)


lukasoppermann:
Hey, thanks for the PR.

Could you please:
1. Create a jsFiddle, Codepen, etc. to demonstrate the problem
2. jQuery is NOT used in the project anymore, so you solution does not work. You would need to revise it using vanilla js and the methods provided by this library

marrouchi:
Done ! https://jsfiddle.net/L247db7h/

lukasoppermann:
Hey, as I mentioned in the email your solution is based on IE special syntax which does not work for us.

Also it think a better solution would be to check if we are on a list with no item, or after the last item, and append it.

lukasoppermann:
Closing due to inactivity.

PR #218, opened 2016-12-09
Revert "Fixed a placeholder bug, and a minor typo"

lukasoppermann:
Reverts voidberg/html5sortable#217

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -1,4 +1,4 @@
-﻿<!doctype html>
+<!doctype html>
 <head>
 	<meta charset="utf-8">
 	<title>HTML5 Sortable library</title>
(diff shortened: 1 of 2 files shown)


thomasmattheussen:
Any reason why this has been reverted? I'm suffering from the same problem as in https://github.com/voidberg/html5sortable/pull/217

lukasoppermann:
Yes, it broke the build. Can you create a ticket for this again? Sorry for the inconvenience. If you could send in a PR with tests that would be great.

lukasoppermann:
Hey @thomasmattheussen, 

I rebuild his PR and added it. Can you please test if your problem is solved? But please make sure to pull from master, it is not tagged yet.

thomasmattheussen:
I fixed it by creating the element and supplying that, instead of a selector. A solution I came up with right after posting this 😅 

But yeah, reverting this once more does seem to fix the original problem :)

PR #221, opened 2017-02-09
Fixes performance issues with firefox

paulius-m:
Hello,
some of our client noticed problems with firefox, then dragging. my investigations showed too much event were passed to onDragOverEnter. debouncing for 25ms worked best.

added _debounce function,
added new option "debounce", default - 0
most part of onDragOverEnter was moved to debounced variant

### DIFF
--- a/src/html.sortable.src.js
+++ b/src/html.sortable.src.js
@@ -438,6 +438,28 @@ var _matches = function(el, selector) {
   return (el.matches || el.matchesSelector || el.msMatchesSelector || el.mozMatchesSelector || el.webkitMatchesSelector || el.oMatchesSelector).call(el, selector);
 };
 
+/**
+ * Creates and returns a new debounced version of the passed function which will postpone its execution until after wait milliseconds have elapsed
+ * @param {fn} Function to debounce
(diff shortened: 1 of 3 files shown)


lukasoppermann:
Thanks, I'll have a look and get back to you.

lukasoppermann:
Hey @paulius-m, firstly thank you very much for contributing. 👍 

I need to ask you to do a couple things before this is ready to be merged.

1. you added `_matches` but there is already an implementation for this [`matches` see here](https://github.com/paulius-m/html5sortable/blob/4bb82946a57c7c0da9ef9bd40928884de99a82d4/dist/html.sortable.js#L53-L71) I have no real preference as to which implementation should be used, yours definitely looks sleeker and less complicated. However I would like you to remove the old `matches` and replace it with your `_matches` so that only one version is in the script.

2. Please do not run `gulp build` and do not commit the files from the dist directory.

3. Please rebase this into 1 commit only with a name like `Fixes performance issues with firefox`

paulius-m:
Hi,
1. I did not add _matches function, i added _debounce.
2. Reverted dist, sorry.
3. OK.

lukasoppermann:
Sorry, I missed that it was there before. Thanks, looks good.

PR #244, opened 2017-04-24
sortable('.sortable', 'disable'); fails

thetaylor82:
Checking the element is draggable on dragstart seems to fix all related issues to the temporary disabling of items.

https://github.com/lukasoppermann/html5sortable/issues/243

### DIFF
--- a/src/html.sortable.js
+++ b/src/html.sortable.js
@@ -552,7 +552,7 @@ var sortable = function (sortableElements, options) {
     // Handle drag events on draggable items
     _on(items, 'dragstart', function (e) {
       e.stopImmediatePropagation()
-      if (options.handle && !_matches(e.target, options.handle)) {
+      if ((options.handle && !_matches(e.target, options.handle)) || !this.draggable) {
         return
       }
 
--- a/src/html.sortable.js
+++ b/src/html.sortable.js
@@ -552,7 +552,7 @@ var sortable = function (sortableElements, options) {
     // Handle drag events on draggable items
     _on(items, 'dragstart', function (e) {
       e.stopImmediatePropagation()
-      if ((options.handle && !_matches(e.target, options.handle)) || !this.getAttribute('draggable')) {
+      if ((options.handle && !_matches(e.target, options.handle)) || this.getAttribute('draggable') === "false") {
         return
       }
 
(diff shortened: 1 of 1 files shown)


lukasoppermann:
Hey thank you for the PR! Is it possible you are wanting to check for `!this.getAttribute('draggable')`? Because currently it fails.

thetaylor82:
Yeah, that's exactly it. Lazy chrome user! I've updated the commit.

lukasoppermann:
Thank you. 👍

jamestwebber:
I have a (I think) related problem: when I disable a sortable element it works fine for the things that are in that element, but the element itself will still accept items from connected sortables.

I'd like to automatically disable it once a certain number of items are added (which might be a useful option for other people), but the 'disable' function only applies to the elements inside rather than the container.

lukasoppermann:
Hey @jamestwebber, I like the idea. Could you send 2 PRs? One for the extension of the disable function to the list itself, so that it does not accept items.

And a second for some kind of `maxItems` value which would not allow more than X items to be added to a list. Could a minItems option also make sense in this content?

jamestwebber:
Happy to write a PR but I'm not really sure where to start (I'm not all that experienced in Javascript). I am not sure how the sortable targets know that they are targets. Any pointers would be appreciated.

lukasoppermann:
Hey @jamestwebber, no worries, I can help you figure it out.

Best would be if you start the first PR about the disabling and we discuss it in there.

The children are gathered here: https://github.com/lukasoppermann/html5sortable/blob/master/src/html.sortable.js#L273

But that should not be necessary for the first PR.

I would suggest to add a `_disabled` property (note the `_` to mark it as private) and set it to `true` here: https://github.com/lukasoppermann/html5sortable/blob/master/src/html.sortable.js#L315 in the `_disableSortable` function.

Than you can check if the sortable is disabled here: https://github.com/lukasoppermann/html5sortable/blob/master/src/html.sortable.js#L684

Don't forget to set `_disabled` to `false` again in the `_enableSortable` function.
https://github.com/lukasoppermann/html5sortable/blob/master/src/html.sortable.js#L287

Just send this as a PR (if you are not sure or have questions, feel free to send a half-done PR as well) and we discuss further steps in your PR. Thank you, looking forward to it.

Issue #249, opened 2017-04-25
Remove jQuery from tests


lukasoppermann:
It is unnecessary and just another dependency. 

Tests need to be cleaned up and jquery selectors, etc removed (this can be done in parts, not all at once).

PR #253, opened 2017-05-08
WIP: _disabled property

jamestwebber:
These are related changes and once I had the first working it was easy to get the second (with caveats I'll get to below).

As discussed in #244, this PR adds a _disabled property so that items from connected sortableElements can't be added. Otherwise, disabling an element only prevents you from picking up items, but not dropping new ones in.

I've also added a new option, `maxSize`, which limits the total number of items. It does this by disabling the element as above.

Two things I couldn't figure out, but would like to add:
1. When you drag an item on top of a disabled sortable, it still shows you the placeholder. Dropping the item there will fail, but ideally it shouldn't show you anything. I think this should be simple but I couldn't parse where in the code the placeholder is displayed.
2. My original intention with `maxSize` was that you'd be able to remove items and re-enable the sortable, but once the sortable is disabled nothing inside of it can be dragged away. From my basic understanding of the code I think changing this would require some reworking of how items and sortable containers are described.
   - thinking about it more, the way I've implemented this is not ideal. You should still be able to _re-order_ the items in a full sortable, you just shouldn't be able to add them. So that part probably needs to change anyway


lukasoppermann:
Thank you @jamestwebber I will look through it later today and give you some feedback. 👍

lukasoppermann:
Hey @jamestwebber,

I had a quick look at it, and it looks pretty nice already.

Here is how I would suggest we continue:

We have a bugfix and a feature, so this should be split into 2 PRs:

The first PR should be the bugfix. Can you also add a test for the bug, so that we can verify it is working when we add new changes? You can look here to see how they might work. https://github.com/lukasoppermann/html5sortable/blob/master/test/events.js (if you need help, le me know).

The second would be the feature you suggested:
The part you added is working fine. However I agree with you that elements should still be sortable inside the element. Users should be able to remove children (thus going below the threshold) and than add new ones.

Disabling the sortable is probably not the right idea here. You would need to check if the list is "full" in the `onDragOverEnter` event function, and if so, `return`.

Also please rename the `maxSize` to `maxItems`. As child-elements are generally referred to as "items" in this package and `maxSize` might be confused with actual pixel-size.

Looking forward to you changes & new PR. Thanks.

jamestwebber:
All right, this is now just a PR for the _disabled flag, and I added some tests. Appears to be working. I'll put the maxItems stuff in a new PR.

(you probably want to do something about the log, I got caught in a bit of git hell trying to sort things out)

edit: Ah dang it merged in the maxSize stuff again, hold on

lukasoppermann:
Hey, thank you for all the work this looks already pretty good.

I think the changes in the main file are done, we just need to tweak the tests. I am not sure what you are trying to achieve here, but think you do not test what you want to test.

I think what you are trying to do is to check if a placeholder is in the sortable, than disable it and check if there is one now. is this correct? Could you maybe explain the idea behind this test a little bit.

Although, if possible can you try to write further tests without jquery? I know, many tests use it, but we will remove it as soon as possible #249 so that would help. 😄 

Apart from the the above mentioned could you please **rebase to master** (let me know if you need help, there will be a small conflict, I believe but it is easy to resolve). And please also remove all files but `src/html.sortable.js` and `test/events.js` from the PR. The dist files only need to be created by me when releasing a new version, otherwise merging is such a hassle. 

Thank you very much so far, looking forward to your reply.

jamestwebber:
I think I have demonstrated that I don't know how to properly fix the Git history for this.

lukasoppermann:
No worries @jamestwebber, I can walk you through it tomorrow. Rebase is not too hard, once you know it, but it is horribly scary in the beginning – been there: 😱

jamestwebber:
I think I got it figured out. As for the tests, I thought those tests were checking to see if an item is added in the right place. I just copied the existing tests and made sure that the indexes did not change. If those tests were testing the placeholders then I must have misread them.

src/html.sortable.js
```diff
@@ -290,6 +290,7 @@ var _enableSortable = function (sortableElement) {
   var items = _filter(_getChildren(sortableElement), opts.items)
   var handles = _getHandles(items, opts.handle)
   _attr(sortableElement, 'aria-dropeffect', 'move')
+  _data(sortableElement, '_disabled', 'false')
```
lukasoppermann, line 293: Why are you not using booleans? Any particular reason?

src/html.sortable.js
```diff
@@ -682,7 +685,7 @@ var sortable = function (sortableElements, options) {
 
     // Handle dragover and dragenter events on draggable items
     var onDragOverEnter = function (e) {
-      if (!dragging || !_listsConnected(sortableElement, dragging.parentElement)) {
+      if (!dragging || !_listsConnected(sortableElement, dragging.parentElement) || _data(sortableElement, '_disabled') === 'false') {
```
lukasoppermann: Wouldn't it need to be `true`?

jamestwebber: Yes that was a mistake brought on my fixing the same accidentally-reverted code over and over...

src/html.sortable.js
jamestwebber: General fear and lack of understanding when it comes to JavaScript crazy type coercion, basically.

Also I didn't look too closely into where _data is stored, the _attr stuff appeared to always be strings.

lukasoppermann: Ahaa, I totally understand that, types are not fun, especially if you come from a strongly typed language. 😁 

Anyway, our `data` has nothing to do with the html data element. It is stored on `$sortable.h5s.data` so types are supported.

src/html.sortable.js
lukasoppermann: Makes sense, have had this many times.

lukasoppermann:
Hey @jamestwebber, I think we should drop the tests for now. They are not testing the right thing and the entire test suit is in a bit of a crazy state 😞. I'll need to clean this up and once this is done it will be easier to write tests.

jamestwebber:
All right that's easy enough to do. If you want to ping me once you've done that refactoring I can try to add them then.

src/html.sortable.js
```diff
@@ -557,7 +560,7 @@ var sortable = function (sortableElements, options) {
     // Handle drag events on draggable items
     _on(items, 'dragstart', function (e) {
       e.stopImmediatePropagation()
-      if ((options.handle && !_matches(e.target, options.handle)) || this.getAttribute('draggable') === 'false') {
+      if ((options.handle && !_matches(e.target, options.handle)) || this.getAttribute('draggable') === 'true') {
```
lukasoppermann: Sorry, I have to bug you one last time. This one was a mistake 😉. You need to keep this as `false` you wanted to change the value below.

src/html.sortable.js
lukasoppermann: This one needs to be `true`.

src/html.sortable.js
jamestwebber: oh whoops!

lukasoppermann:
Hey @jamestwebber, I will gladly ping you once the tests are in a better state. I think you made a little mistake and changed the wrong value from false to true (see comments). Once this is changed I can merge it.

Thank you very much. Keep the PRs coming, looking forward to your help. 👍

PR #257, opened 2017-05-08
added maxItems option

jamestwebber:
Here is the PR for maxItems (alone). I didn't write a test for this, I am not sure where the best place to put it would be. There are currently no explicit tests of connectWith that I could see, and that would be the starting point.


lukasoppermann:
Hey @jamestwebber, thank you very much, I will have a look at it now.

Are you saying you would need a connected test to know how to do it, or are you just not sure where to put it? If its the location, just create a new file: `connectedSortable.js` in the test folder. Also please try to not use jQuery, as mentioned in the other ticket we are removing it from the test.

Also can you please to the following for me:

- rebase to master
- remove all commits from your other PR in this one (use `git rebase -i HEAD~N` where N is the number of commits you added) now you can drop commits by replacing the `pick` with a `drop` (sorry, if you know this already, just making sure)
- please remove all files but `src/html.sortable.js` from the PR
- please adjust the `Readme.md` to mention your new feature and how it can be used

Thanks. 👍

src/html.sortable.js
```diff
@@ -682,7 +692,10 @@ var sortable = function (sortableElements, options) {
 
     // Handle dragover and dragenter events on draggable items
     var onDragOverEnter = function (e) {
-      if (!dragging || !_listsConnected(sortableElement, dragging.parentElement)) {
+      if (!dragging || !_listsConnected(sortableElement, dragging.parentElement) || _data(sortableElement, '_disabled') === 'true') {
+        return
+      }
+      if (maxItems && _filter(_getChildren(sortableElement), _data(sortableElement, 'items')).length >= maxItems) {
```
lukasoppermann, line 698: You need to do `....length - 1` otherwise if I set `maxItems` to 3, I can only add 2 items.

src/html.sortable.js
```diff
@@ -554,6 +559,11 @@ var sortable = function (sortableElements, options) {
       })
     }
 
+    // max items
+    if (options.maxItems && typeof options.maxItems === 'number') {
```
lukasoppermann, line 563: Couldn't you just remove this entire `if` block and use `options.maxItems`in line 698? You could check for typeof number down there.

src/html.sortable.js
jamestwebber: This is because of the placeholder class getting added. If you pass an "items" option that doesn't match the placeholder, it works correctly this way. I think that's better than having it always off-by-one which seems confusing.

src/html.sortable.js
jamestwebber: Sure, I don't know if a type check is even necessary, just sort of emulating what I saw elsewhere. The type check probably isn't necessary at all, it will act weird if people give a weird parameter but that's on them.

jamestwebber:
Uh, [that probably wasn't the right thing to do](https://xkcd.com/1597/)...I have a branch called maxitems with the correct history and changes but it's no longer attached to this PR.

lukasoppermann:
Hey @jamestwebber, no worries just create a new PR with the branch. (Or is there another issue I don't understand. )

PR #260, opened 2017-05-10
Maxitems

jamestwebber:
New PR to replace #257 because I borked the Git log

### DIFF
--- a/src/html.sortable.js
+++ b/src/html.sortable.js
@@ -471,7 +471,7 @@ var sortable = function (sortableElements, options) {
       draggingClass: 'sortable-dragging',
       hoverClass: false,
       debounce: 0,
-      maxItems: null
+      maxItems: 0
     }
     for (var option in options) {
       result[option] = options[option]
--- a/src/html.sortable.js
+++ b/src/html.sortable.js
@@ -470,7 +470,8 @@ var sortable = function (sortableElements, options) {
       placeholderClass: 'sortable-placeholder',
       draggingClass: 'sortable-dragging',
       hoverClass: false,
-      debounce: 0
+      debounce: 0,
+      maxItems: null
     }
     for (var option in options) {
       result[option] = options[option]
@@ -510,6 +511,7 @@ var sortable = function (sortableElements, options) {
     var index
     var startParent
     var placeholder = options.placeholder
+    var maxItems
     if (!placeholder) {
       placeholder = document.createElement(
         /^ul|ol$/i.test(sortableElement.tagName) ? 'li' : 'div'
@@ -552,6 +554,11 @@ var sortable = function (sortableElements, options) {
       _on(items, 'mouseleave', function () {
         this.classList.remove(hoverClass)
       })
+    }
+
+    // max items
+    if (options.maxItems && typeof options.maxItems === 'number') {
+      maxItems = options.maxItems
     }
 
     // Handle drag events on draggable items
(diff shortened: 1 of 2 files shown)


jamestwebber:
Still needs a test of a pair of connected sortables

src/html.sortable.js
```diff
@@ -685,6 +692,9 @@ var sortable = function (sortableElements, options) {
       if (!dragging || !_listsConnected(sortableElement, dragging.parentElement)) {
         return
       }
+      if (maxItems && _filter(_getChildren(sortableElement), _data(sortableElement, 'items')).length >= maxItems) {
```
lukasoppermann, line 695: I get your point: https://github.com/lukasoppermann/html5sortable/pull/257 (you talking about "class" confused me, I think it means something else outside of js). The clean solution would be to add a new function which filters children differently to only include items with a sortable-item-id.

Until we have this working, can you please add the `-1` thingy and a comment `//TODO: remove -1 once placeholder items are not included when getting children`.

Otherwise people will have a bug.

src/html.sortable.js
```diff
@@ -470,7 +470,8 @@ var sortable = function (sortableElements, options) {
       placeholderClass: 'sortable-placeholder',
       draggingClass: 'sortable-dragging',
       hoverClass: false,
-      debounce: 0
+      debounce: 0,
+      maxItems: null
```
lukasoppermann: this should be `0` no `null`.

lukasoppermann:
Hey @jamestwebber, looks great. Added some small comments. 

Are you still looking into writing tests or would you rather we create deal with them later?

jamestwebber:
If you could write the tests that would be great, I don't think I can figure out the right structure while also removing jQuery usage.

src/html.sortable.js
jamestwebber: I think that would cause it's own bugs in other cases? If you have an element that's a child of your sortable but not supposed to be draggable. Maybe that's not a realistic use-case though.

lukasoppermann: Hey I don't really understand how this would be a problem, can you explain it in more detail? I don't get what you mean. 😄

lukasoppermann:
Hey @jamestwebber, sure, once we have the -1 or not part figured out I'll merge it without tests. One of us can write the tests later on.

Maybe you find some more enhancements or bugs to fix were writing tests is easier (personally I find it really hard to write tests for UI elements where user interaction is required for the test).

src/html.sortable.js
jamestwebber: We want to know how many items the sortable contains. If we just count the children of the sortable, we'll count the placeholder and anything else that's in there at the time&mdash;we don't know how many non-"items" are inside the sortable, so no constant offset will work in all cases.

If we filter the children by the `items` option we don't need an offset, but it only works if the user specified that option in a way that excludes the placeholder and anything else inside the sortable. This could be explicitly mentioned in the documentation.

I think your solution of filtering for sortable-item-id is the right thing to do, and lacking that change having no offset (and mentioning this in the doc) is second-best.

lukasoppermann: Hey, I am fine with leaving it as long as we mention it in the docs. Could you add this to your PR? Thanks.

lukasoppermann:
Thanks @jamestwebber 👍

PR #272, opened 2017-06-24
Update GitHub URL

nazar-pc:
Old URL still works and redirects properly, but since it was changed it should better be updated here as well


lukasoppermann:
Thanks, go one spotting it. 👍

PR #275, opened 2017-08-09
Firefox bug fix

DanTheOrange:
Fixed bug for Firefox where data needs to have content in order for html drag and drop to function.

### DIFF
--- a/src/html.sortable.js
+++ b/src/html.sortable.js
@@ -173,7 +173,9 @@ var _removeSortableEvents = function (sortable) {
 var _attachGhost = function (event, ghost) {
   // this needs to be set for HTML5 drag & drop to work
   event.dataTransfer.effectAllowed = 'move'
-  event.dataTransfer.setData('text', '')
+  // Firefox requires some arbitrary content in the data in order for
+  // the drag & drop functionality to work
+  event.dataTransfer.setData('text', 'arbitrary-content')
 
   // check if setDragImage method is available
   if (event.dataTransfer.setDragImage) {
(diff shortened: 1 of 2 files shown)


DanTheOrange:
I have made this change locally and tested it in Firefox and the latest Chrome. I don't have access to IE or Edge to make sure it doesn't break but I don't believe it will make a difference to them. Here is an archaic stack overflow post for reference https://stackoverflow.com/questions/3977596/how-to-make-divs-in-html5-draggable-for-firefox

Thanks :)

lukasoppermann:
Hey @DanTheOrange,

thanks for the PR. I see no reason why it should fail on IE. I will merge it tomorrow.

Jeremy-F:
Hey, @DanTheOrange & @lukasoppermann 

Thank's for this update, but there is still a Firefox bug with the handler. 

Let’s see : https://jsfiddle.net/k8nvwapa/3/

Thanks for all that work

lukasoppermann:
Hey @Jeremy-F, I think the issue is, that you are using a button element, which cancels the drag because of its click event. Try this one: https://jsfiddle.net/k8nvwapa/4/

Jeremy-F:
Thanks again @lukasoppermann 
It's crazy that it works on Chrome/Safari & other but Firefox.

DanTheOrange:
I was having this issue too. I found the reason on caniuse as this was the original bug I was fixing on my project.

"In Firefox, the dragstart event does not fire on button elements. This effectively disables drag and drop for button elements." https://caniuse.com/#feat=dragndrop in the known issues section.

lukasoppermann:
@DanTheOrange thank you, added this note to the readme in #279

PR #290, opened 2017-12-07
Update index.html

NguyenDa18:
I cloned the repo and it looks like since the last update the size of html.sortable.min.js is now 7KB as opposed to 2KB as advertised. Still a very impressive feat! Using the code to make a really lightweight website.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -27,7 +27,7 @@
 		    <div class="sm-flex flex-center mxn2">
 		      <div class="flex-auto px0 py2">
 						<ul>
-							<li>Only 7KB (minified and gzipped).</li>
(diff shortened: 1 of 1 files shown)


NguyenDa18:
7kb

lukasoppermann:
Hey, actually if you gzip the file, it has by now grown to 2.54KB. However the 7KB you are referring to is without gzip.

If you could update it to 2.5KB that would be great and I can merge it. Thanks.

NguyenDa18 (commented):
Ohhh I guess I read over the g-zip portion, change made as you requested.

lukasoppermann:
Thanks.

PR #295, opened 2018-02-14
added 'acceptFrom' option to allow connection lists more flexibly

mfriesen:
I have 2 lists. Where list 1 can be sorted into list 2, but list 1 cannot be sorted into list 1. I've added a listconnected option to allow custom rules when lists can be sorted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,14 @@ sortable('.sortable', {
   maxItems: 3 // Defaults to 0 (no limit)
 });
 ```
+### acceptFrom
+Use the `acceptFrom` option to restrict which sortable's items will be accepted by this sortable. acceptForm accepts a space separeted list of selectors or false to accept nothing. This is an alternative to #connectwith and should not be used together.
+
+``` javascript
+sortable('.sortable', {
+  acceptForm: '.selector,.anotherSortable' // Defaults to null
+});
+```
 
 ## Methods
 
--- a/docs/index.html
+++ b/docs/index.html
@@ -267,10 +267,64 @@
           </div>
       </div>
     </section>
+
+	<section class="mb3 mx-auto col col-12">
+  		<div class="p3 clearfix bg-navy yellow">
+				<div class="col col-6">
+			    <h2 class="h3 m0">
+						acceptFrom
+			    </h2>
+			     <p>acceptFrom allows to control which sortable to accept  drops from.</p>
+			     <p>Item 1-6 are allows to be sorted with Item 7-12 but not the other way around.</p>
+					<div class="mt2 p2 bg-navy border yellow border-yellow">
+				    <code class="mb0">
+							<div>sortable('.o-sortable1', {</div>
+							<div>acceptFrom: false (accept from no one)</div>
+							<div>});</div>
+							<div>sortable('.o-sortable2', {</div>
+							<div>acceptFrom: '.o-sortable1' (accept from .o-sortable1)</div>
+							<div>});</div>
+						</code>
+  				</div>
+		    </div>
+			    <div class="col col-3">
+						<ul class="ml4 js-sortable-oneway sortable list flex flex-column list-reset">
+							<li class="p1 mb1 navy bg-yellow" style="position: relative; z-index: 10">Item 1</li>
+							<li class="p1 mb1 navy bg-yellow" style="position: relative; z-index: 10">Item 2</li>
+							<li class="p1 mb1 navy bg-yellow" style="position: relative; z-index: 10">Item 3</li>
+							<li class="p1 mb1 navy bg-yellow" style="position: relative; z-index: 10">Item 4</li>
+							<li class="p1 mb1 navy bg-yellow" style="position: relative; z-index: 10">Item 5</li>
+							<li class="p1 mb1 navy bg-yellow" style="position: relative; z-index: 10">Item 6</li>
+							<li class="mb1 bg-navy ghost border border-yellow" style="position: relative; z-index: 1; margin-top: -49px; height: 40px; display: block;">Ghost test</li>
+						</ul>
+			    </div>
+				<div class="col col-3">
+					<ul class="ml4 js-sortable-oneway-receive sortable list flex flex-column list-reset">
+						<li class="p1 mb1 navy bg-yellow" style="position: relative; z-index: 10">Item 7</li>
+						<li class="p1 mb1 navy bg-yellow" style="position: relative; z-index: 10">Item 8</li>
+						<li class="p1 mb1 navy bg-yellow" style="position: relative; z-index: 10">Item 9</li>
+						<li class="p1 mb1 navy bg-yellow" style="position: relative; z-index: 10">Item 10</li>
+						<li class="p1 mb1 navy bg-yellow" style="position: relative; z-index: 10">Item 11</li>
+						<li class="p1 mb1 navy bg-yellow" style="position: relative; z-index: 10">Item 12</li>
+						<li class="mb1 bg-navy ghost border border-yellow" style="position: relative; z-index: 1; margin-top: -49px; height: 40px; display: block;">Ghost test</li>
+					</ul>
+		    </div>
+			</div>
+		</section>    
 	</div>
   <!-- <script src="../src/html.sortable.js"></script> -->
   <script src="html.sortable.min.js"></script>
 	<script>
+	sortable('.js-sortable-oneway', {
+      forcePlaceholderSize: true,
+      acceptFrom: false,
+      placeholderClass: 'p1 mb1 bg-navy border border-yellow',
+    });
+    sortable('.js-sortable-oneway-receive', {
+      forcePlaceholderSize: true,
+      acceptFrom: '.js-sortable-oneway,.js-sortable-oneway-receive',
+      placeholderClass: 'p1 mb1 bg-navy border border-yellow',
+    });
     sortable('.js-sortable', {
       forcePlaceholderSize: true,
       placeholderClass: 'p1 mb1 bg-navy border border-yellow',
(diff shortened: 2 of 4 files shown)


lukasoppermann:
Awesome, I will have a look later on and merge/get back to you. Thanks.

lukasoppermann:
Hey @mfriesen,

looks interesting. Could you please show me the function you use for your use case, so I can understand it better?

I don't really understand why your test makes sense, can you explain it? It looks like to does not really test anything to me.

Also can you please change the naming? `f` is not a very good variable for somebody studying the code. Maybe just call it `listConnectedCallback`. And please rename `listconnected` to `listConnected` to stay inline with the current naming convention we use. 

Thank you very much. Looking forward to merging this.

mfriesen:
Hey @lukasoppermann 

I'll update the docs/index.html to show an example on how this would be used.

I'm not familiar with JavaScript testing, so I may have done it wrong. What I was trying to do is show that when using a listConnectedCallback that always returns false, objects that were previous connected, IE the same object are no longer connected. If you have suggestions on how to make it better, please let me know.

I will update the naming convention as per your suggestion.

mfriesen:
@lukasoppermann after reviewing, maybe a callback isn't the best approach. It makes things more complicated. I am thinking of adding a dropDisable option instead which would disable all drops on a sortable. 
It would be similar to the .disabled examples you created except it applies to the entire sortable list.

lukasoppermann:
Hey @mfriesen yeah, I see your point. But maybe it would be better to improve the current connectWith method to allow one way connections? That would solve the same issue, right?

Or maybe to keep bc we could add something like `acceptFrom` where you can put a `space-separated` list of classes or something. What do you think about that?

mfriesen:
@lukasoppermann I really like the acceptFrom idea. Could default it to null to accept everything, empty string means except nothing and a space-separated list. That should work for me.

mfriesen:
@lukasoppermann I've added the acceptFrom. Let me know what you think.

lukasoppermann:
Hey @mfriesen, thanks for the update. My idea was actually different:

Your implementation:
- user connects lists with connectWith
- user uses acceptFrom to restrict connection

My idea:
- acceptFrom is basically a replacement for connectWith (we just leave both in for now to allow for backwards compatibility)
- user can ignore connectWith and just use acceptFrom
- if acceptFrom is null or false, the list does not accept any items
- if it has one or more list items in a space separated list / or an array (depending how we do it with other things in the plugin) items from those lists are accepted.

Do you understand the difference or am I explaining it badly? 😉  What do you think about this?
Do you agree or if you prefer your idea, can you explain to me why you think it is better? 

I currently think ignoring `connectWith` and phasing it out later on makes it easier and less work intense for the user. Correct me if I am wrong or if there is something that we cannot do without connectWith.

Thank you for your work and quick response, very awesome! 👍

mfriesen:
@lukasoppermann sorry I misunderstood. Yes I like that better. I'll fix it.

mfriesen:
@lukasoppermann is this what you are thinking? `.list1` accepts no drops and `.list2` accepts drops from `.list1`.

```js
sortable('.list1', { acceptFrom: '' });
sortable('.list2', { acceptFrom: '.list1 .anotherlist' });
```

lukasoppermann:
@mfriesen yes. 

Although I would rather generally say you would do

```js
sortable('.list1', { acceptFrom: false });
sortable('.list2', { acceptFrom: '.list1 .anotherlist' });
````

or probably even

```js
sortable('.list1');
sortable('.list2', { acceptFrom: '.list1 .anotherlist' });
```

So the default value should be `false` (which would mean by default the behaviour would be like the current behaviour).

Of course if you do, do `acceptFrom: ''` it would convert it to an array of nothing, which would end up having the same result as `false`.

So yes. Do you get what I mean, or was that just confusing gibberish?

mfriesen:
@lukasoppermann I've updating the pull request. Let me know if I got it right or wrong.

README.md
```diff
@@ -186,6 +186,14 @@ sortable('.sortable', {
   maxItems: 3 // Defaults to 0 (no limit)
 });
 ```
+### acceptFrom
+Use the `acceptFrom` option to restrict the which sortable's items will be accepted by this sortable. acceptForm accepts a space separeted list of selectors or '' to accept nothing.
```
lukasoppermann: - restrict **the** which (the should be removed)
- `or '' to accept nothing` I would rather use `null` instead of '' as it is more explicit for the user.
- maybe it would be good to say that this is an alternative to `connectWith` and should not be used together

README.md
```diff
@@ -186,6 +186,14 @@ sortable('.sortable', {
...
+
+``` javascript
+sortable('.sortable', {
+  acceptForm: '.selector' // Defaults to false (defaults to connectWith behavior)
```
lukasoppermann: Maybe it would be better to use `'.selector .anotherSortable'` to show people how it works with multiple items (some users are very new to js).

Also looking at it now, do you think a comma separated list would be better? My reason being that with space-separated lists you can not use nested selectors. (Maybe it is not needed though, what is your thought on it).

docs/index.html
```diff
@@ -267,10 +267,64 @@ <h2 class="h3 m0">
...
+		    </div>
+			</div>
+		</section>    
 	</div>
   <!-- <script src="../src/html.sortable.js"></script> -->
   <script src="html.sortable.min.js"></script>
 	<script>
+	sortable('.js-sortable-oneway', {
+      forcePlaceholderSize: true,
+      acceptFrom: '',
```
lukasoppermann: I'd rather also use `null` here instead of empty string.

lukasoppermann:
Hey @mfriesen very awesome, thank you! 👍 
I added some tiny comments (I think they are nearly all the '' vs. `false`). If you could fix them it should be ready to merge. I think this is a very clean and simple way of implementing it, nice job.

mfriesen:
@lukasoppermann I don't see your comments. Where can I find them?

lukasoppermann:
I think I forget to submit, can you seem then now?

README.md
mfriesen: My initial thoughts were comma separated. I'll change it to comma separated.

mfriesen:
@lukasoppermann The problem with using acceptFrom: false, to disable drops, is that is the same as the default option value. So setting it to false will skip the acceptFrom function and run
```
if (curList === destList) {
    return true
  }
```
Which will disable drops from other sortables but not from it's self. If I change the acceptFrom default value to null. That would fix the problem.  Thoughts?

lukasoppermann:
Hey @mfriesen I don't really understand the issue, sorry. 

> The problem with using acceptFrom: false, to disable drops, is that is the same as the default option value. So setting it to false will skip the acceptFrom function and run […] Which will disable drops from other sortables but not from it's self.

Isn't this exactly the behavior we want? Or are you saying setting it to `null` will also disable dropping on itself and this is a desired effect?

If so, is a use case having a list of items and doing copy on drag? If so, what happens if I start dragging and than let go of the item?

mfriesen:
Right now the default options are
```
options = (function (options) {
    var result = {
      acceptFrom: false,
```

And in _listsConnected I do the acceptFrom if
```
if (curOpts.acceptFrom !== false) {
```
So if you create a sortable like
```
sortable('.sortable', {
      acceptFrom: false,
});
```
This will skip the curOpts.acceptFrom !== false check and since the next line is
```
 if (curList === destList) {
    return true
  }
```
It will accept from its self. I thought acceptFrom:false should accept from nothing. Therefore if we change the options to default as null and then change
```
if (curOpts.acceptFrom !== null) {
```
Then setting acceptFrom:false will accept from no one. Which is what I thought we want.

README.md
lukasoppermann: 👍

lukasoppermann:
Hey @mfriesen yes, I agree, `false` should be not accepting anything your are right, that totally makes sense. 👍
Can we than change it to `null` instead of empty string `''`.

Also: Would you agree that after this is merged we can deprecate (not remove, but mark as deprecated) the `connectWith` feature?

test/internal.js
```diff
@@ -134,6 +134,22 @@ describe('Internal function tests', function () {
...
     // test if sortable2 is connected to sortable3 (should be false)
     assert.equal(window.sortable.__testing._listsConnected(connectedUl, notConnectedUl), false)
+
+    window.sortable(connectedUl, 'destroy')
+    window.sortable(notConnectedUl, 'destroy')
+    window.sortable(connectedUl, {
+      acceptFrom: '.sortable3'
+    })
+    window.sortable(notConnectedUl, {
+      acceptFrom: ''
```
lukasoppermann: So for this one we should probably test both `''` empty string and `null`.

README.md
lukasoppermann: But the comment would have to be `// Defaults to null` correct? And I think we should remove the `(defaults to connectWith behavior)` part here, as it might be confusing.

mfriesen:
@lukasoppermann I've updated the code as per your suggestions. Let me know if I got all the changes right.

src/html.sortable.js
```diff
@@ -242,6 +242,12 @@ var _removeItemData = function (items) {
  * @param {Element} destList
  */
 var _listsConnected = function (curList, destList) {
+  var acceptFrom = _data(curList, 'opts').acceptFrom
+  if (acceptFrom !== null) {
+    return acceptFrom !== false && acceptFrom.split(',').filter(
+      sel => sel.length > 0 && _matches(destList, sel)
```
lukasoppermann: One last thing, can you please convert this to an oldschool function:
```
      function (sel) {
        return sel.length > 0 && _matches(destList, sel)
      }
```

Otherwise it throws of the uglify (which would probably be solved with an update) but it also is not supported in IE11

lukasoppermann:
Hey @mfriesen I have one last tiny comment, the rest is superb. 😉

mfriesen:
@lukasoppermann I've updated the function.

lukasoppermann:
Perfect, thank you. 👍

Issue #265, opened 2017-05-13
Copy option


lukasoppermann:
Allow for items to be `copied` (cloned) instead of moved to a connected list.

WoodyAlmeida:
Was looking exactly for this, unfortunately I don't have enough experience in JS to do it. Anybody knows if someone is working on this?

lukasoppermann:
Hey @WoodyAlmeida, sadly I don't think so. I know you said you don't have that much js experience, but if you would like to send in a PR I can help you with it. Basically you would need to allow for a new option `copy: true` passed into the sortable function. 

You would need to check it when dragging and only remove the old item, if copy is set to false.

Best would be to instead of checking the copy option directly, creating a function `isCopyActive` which returns `opts.copy` for now but could later be enhanced to check for keyboard shortcuts.

mfriesen:
I have added this copy feature. I will be creating a pull request soon.

lukasoppermann:
Great initiative @mfriesen, feel free to also send a WIP PR so we can discuss stuff early on.

mfriesen:
@lukasoppermann The change doesn't seem very large.
1) In the 'dragstart' event where "dragging = this". 
I changed it (if the "copy:true" option is set) to return this.cloneNode(true).

One issue I ran into was if you have an unsuccessful drop (IE: You drop it on a list where _listsConnected returns false).  The copied element needs to be removed from the DOM in the 'dragend' event.

So, I added "_data(dragging, '_dropped', 'true')" into the 'drop' event so I know if the element was successfully dropped in the 'dragend' event. That way I can remove the element if the drop failed.

any suggestions?

lukasoppermann:
@mfriesen hey, can you send a PR? It sounds good, but I find it easier to judge if I see the code. 😁 

Also please try to add some tests (you can send a PR first without any tests though and add them later) to verify: cloning works, deleting works, and drop only works for allowed items and the that sortupdate event fires correctly and the data in it makes somewhat sense. 👍  thank you.

lukasoppermann:
Hey @mfriesen couldn't you actually use the `copy` keyword in the `dropEffect` instead of cloning the node? https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/dropEffect

mfriesen:
I believe that only changes the visual cursor. But I believe the cloning is still required.
[...]

PR #301, opened 2018-02-20
Added Copy Feature

mfriesen:
Added Copy option for elements instead of move, as per issue #265 
```
sortable('.o-sortable', {
copy: true
}
```


lukasoppermann:
Hey @mfriesen, thanks a lot, this looks pretty good. I will have a closer look later today or tomorrow (traveling today). Could you do me a favor and rebase to master? (I did some minor tweaks to the readme after I merged your PR).

mfriesen:
I've rebased from master.

src/html.sortable.js
```diff
@@ -256,6 +275,13 @@ var _listsConnected = function (curList, destList) {
   }
   return false
 }
+/*
+ * Is Copy Active for sortable
+ * @param {Element} sortable a single sortable
+ */
+var _isCopyActive = function (sortable) {
+  return _data(sortable, 'opts').copy
```
lukasoppermann: This would need to be `return _data(sortable, 'opts').copy === true` because otherwise e.g. setting copy to 'no' would also evaluate to true.

src/html.sortable.js
```diff
@@ -662,6 +695,8 @@ var sortable = function (sortableElements, options) {
 
       e.preventDefault()
       e.stopPropagation()
+
+      _data(dragging, '_dropped', 'true')
```
lukasoppermann: Why are you using `_dropped` and not `dropped`?

README.md
```diff
@@ -197,7 +197,22 @@ sortable('.sortable', {
   maxItems: 3 // Defaults to 0 (no limit)
 });
 ```
+### acceptFrom
```
lukasoppermann: Somehow your rebase did not work, this should not be in your PR `acceptFrom` and I changed and moved it as well (see the current readme). Can you please try to rebase again (and pull master first. 😉in case you forgot). Thanks.

lukasoppermann (changes requested):
Really nice, just some tiny things. I don't really like adding `_data(dragging, '_dropped', 'true')` for the sole reason of deleting it later on, but I also don't have a better idea at the moment. 😆 

Thanks.

mfriesen:
@lukasoppermann I should have made all the fixes.

lukasoppermann (approved):

Issue #296, opened 2018-02-16
Change Item on Drag


mfriesen:
Similar to issue #265. I would like to change the item being dragged. IE: When dragging a &lt;li&gt; item, change it to a <input />. 

I was thinking of a adding a copyFrom option where the copyFrom is the ID of an hidden element? I only need it when the copy=true. I adding this option when the copy=false would be tricky as you would have to swap the item back, if the drop failed.

lukasoppermann:
Hey, so you are saying, you have for e.g. a some draggable `<li>s` and on drag want to change it to an `<img>`? 

This sounds like a very specific case/edge case to me. Maybe it would be better to somehow do this in the `onDragStart` event? You could check if the `copy` option is set to true and than do it, correct? Do you think that would be a good solution as well?

If so, maybe you can add a `recipes` sections or something to the readme and add this example? 

I just want to make sure we don't add to many things and complicate the whole package. 😉 
However feel free to provide some arguments why it would be better to add this as a specific option and we can discuss it.

lukasoppermann:
@mfriesen do you think your issue is the same as #289?

mfriesen:
Yes, I believe so. Basically I'm creating a toolbar and when you drag the toolbar item, I need it to expand as the type of item it is dropping. I will try and see if onDragStart will work.

mfriesen:
onDragStart is the right time but I ran into a problem. The sortstart only fires if the source sortable is connected to the drop sortable. IE: If list1 is acceptFrom: false and list2 is acceptFrom: .list1. The event will not fire. I am not entirely sure why?

For example in the acceptFrom example. If you add
```
sortable('.js-sortable-oneway')[0].addEventListener('sortstart', function(e) {
    	console.log("sortstart....")
	});
```
This will never fire.

lukasoppermann:
Hmm, this seems to be because of `_dispatchEventOnConnected` https://github.com/lukasoppermann/html5sortable/blob/master/src/html.sortable.js#L433-L438 although I don't see the sense in it. I guess we should carefully remove it.

I would guess this function is used so that a non-connected sortable does not get the event if I drop something onto it (which will not work, because it is not connected). However, this does not really make sense for `sortstart` I think.

mfriesen:
I agree. I think creating a _dispatchEvent function and calling that from sortstart would be fine and I don't think it would break anything. Let me know if you want me to help out.

lukasoppermann:
[...]

PR #302, opened 2018-02-22
Fix sortstart event to fire everytime

mfriesen:
The sortstart event only fired if the lists were connected or the same, but won't fire across 2 different lists or if they are not connected. see #296

### DIFF
--- a/src/html.sortable.js
+++ b/src/html.sortable.js
@@ -16,7 +16,6 @@
 var dragging
 var draggingHeight
 var placeholders = []
-var sortables = []
 /**
  * Get or set data on element
  * @param {Element} element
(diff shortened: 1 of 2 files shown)


src/html.sortable.js
```diff
@@ -610,7 +618,7 @@ var sortable = function (sortableElements, options) {
       startParent = this.parentElement
       startList = _serialize(startParent)
       // dispatch sortstart event on each element in group
-      _dispatchEventOnConnected(sortableElement, _makeEvent('sortstart', {
+      _dispatchEvent(sortableElement, _makeEvent('sortstart', {
```
lukasoppermann: Sorry, but looking at it now, couldn't we just do `sortableElement.dispatchEvent( _makeEvent('sortstart', {…` instead if using the `_dispatchEvent` function? Sorry, I did not catch that before you wrote it.

Also the comment should be updated

mfriesen:
@lukasoppermann I've fixed it and I've changed sortupdate & sortstop to match. I don't believe this will break backwards compatibility

lukasoppermann:
Hey @mfriesen did you rebase to master? Is it possible that you forgot to push the fixes? Because it looks the same to me.

mfriesen:
@lukasoppermann my bad. It's up now.

lukasoppermann:
Hey, the question with removing them is, will `sortupdate` & `sortstop` be called on the not connected sortables if dropped on them? Did you check this? Could you add tests for this as well? Basically the same tests with not connected, which should turn out to be false on the called variablesß

Thank you.

mfriesen:
@lukasoppermann ok, I think I fixed it up much better. The real problem is that the events don't fire when acceptFrom is used.
You'll also noticed I removed firing of the 'dropend' event from the 'drop' event. The browser does this already, so this didn't actual do anything.

lukasoppermann:
Thank you, looks good! :+1:

PR #321, opened 2018-02-26
Refactor sortable events

### LINKED ISSUE #311: Fix Copy Option
Hey @mfriesen there is an issue with the copy option:
https://lukasoppermann.github.io/html5sortable/index.html

- [ ] The new items that are added to a sortable are not part of it, so we would need to reinit every time a copy is added to an item. (This is not so good, maybe we could fix this by fixing the general issue: adding new items to a sortable should automatically work, without the reinit: #312)
- [ ] A bug in copy is that when I move a copied item onto an invalid target (and my copy start list has `acceptFrom:false`) the item still goes back to the parent. This could also be seen in a bigger picture so that dropping outside a valid target should cancel the sorting action (similar to when pressing `esc` #313)

mfriesen:
@lukasoppermann In order to fix #311. I need to refactor the initializing of the sortable so I can initialize the copy element.  Let me know what you think about these changes.

### DIFF
--- a/src/html5sortable.ts
+++ b/src/html5sortable.ts
@@ -16,6 +16,10 @@
 var dragging
 var draggingHeight
 var placeholders = []
+var placeholderMap = new Map();
+var index
+var startParent
+var startList
 /**
  * Get or set data on element
  * @param {Element} element
@@ -84,10 +88,12 @@ var _on = function (element, event, callback) {
     }
     return
   }
-  element.addEventListener(event, callback)
-  element.h5s = element.h5s || {}
-  element.h5s.events = element.h5s.events || {}
-  element.h5s.events[event] = callback
+  if (!(element.h5s && element.h5s.events && element.h5s.events[event])) {
+    element.addEventListener(event, callback)
+    element.h5s = element.h5s || {}
+    element.h5s.events = element.h5s.events || {}
+    element.h5s.events[event] = callback
+  }
 }
 /**
  * @param {Array|Element} element
@@ -548,11 +554,7 @@ export default function sortable (sortableElements, options) {
     _reloadSortable(sortableElement)
     // initialize
     var items = _filter(_getChildren(sortableElement), options.items)
-    var index
-    var startParent
-    var startList
     var placeholder = options.placeholder
-    var maxItems
     if (!placeholder) {
       placeholder = document.createElement(
         /^ul|ol$/i.test(sortableElement.tagName) ? 'li' : 'div'
@@ -566,6 +568,7 @@ export default function sortable (sortableElements, options) {
 
     _data(sortableElement, 'items', options.items)
     placeholders.push(placeholder)
+    placeholderMap.set(sortableElement, placeholder)
     if (options.acceptFrom) {
       _data(sortableElement, 'acceptFrom', options.acceptFrom)
     } else if (options.connectWith) {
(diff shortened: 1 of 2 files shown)


src/html5sortable.ts
```diff
@@ -16,6 +16,10 @@
 var dragging
 var draggingHeight
 var placeholders = []
+var placeholderMap = new Map();
+var index
```
lukasoppermann, line 20: Did you test that moving them outside the forEach still works even if you have multiple items? (I think it should work, but just making sure).

src/html5sortable.ts
```diff
@@ -84,10 +88,12 @@ var _on = function (element, event, callback) {
     }
     return
   }
-  element.addEventListener(event, callback)
-  element.h5s = element.h5s || {}
-  element.h5s.events = element.h5s.events || {}
-  element.h5s.events[event] = callback
+  if (!(element.h5s && element.h5s.events && element.h5s.events[event])) {
```
lukasoppermann, line 91: I think this does not work. When an item gets some data via the `_data` function it is added to `element.h5s` or when an element already has an event, this way you could not add another event, because `element.h5s` and `element.h5s.events` already exist.

test/events.js
```diff
@@ -230,7 +230,7 @@ describe('Testing events', function () {
     assert.equal(event.dataTransfer.dropEffect, 'move')
 
     event = window.sortable.__testing._makeEvent('drop')
-    // Object.defineProperty(event, 'target', {value: secondLi, enumerable: true})
+    Object.defineProperty(event, 'target', {value: secondLi, enumerable: true})
```
lukasoppermann, line 233: I think this is wrong. The target will be the placeholder, not the item. Please verify and correct me if I am wrong, but when I tested I think that the target was the placeholder element.

src/html5sortable.ts
```diff
@@ -577,185 +580,225 @@ export default function sortable (sortableElements, options) {
...
 
     // Mouse over class
-    if (options.hoverClass) {
-      var hoverClass = 'sortable-over'
-      if (typeof options.hoverClass === 'string') {
-        hoverClass = options.hoverClass
-      }
+    var hoverClass = _getHoverClass(sortableElement)
+    if (hoverClass != null) {
+      _on(items, 'mouseenter', mouseenter)
```
lukasoppermann, line 585: I think it would be better to use anonymous functions here because by using named functions you duplicated getting hoverClass and checking it for `null`.

Using an anonymous function would remove this duplication and make it easer to read.

By the way, since rollup now transpiles to es5 you can use fat arrow functions, if you want.

src/html5sortable.ts
```diff
@@ -566,6 +568,7 @@ export default function sortable (sortableElements, options) {
 
     _data(sortableElement, 'items', options.items)
     placeholders.push(placeholder)
+    placeholderMap.set(sortableElement, placeholder)
```
lukasoppermann, line 571: Why are you adding the Map? Is it to more easily get the placeholder? Is this necessary for this PR? Otherwise how about a different PR where you replace `placeholders` with a map? Would that work? I really like the idea of using the map, but it currently feels like it doubles a bit with `placeholders`. Please let me know if I am wrong/missing something.

lukasoppermann (changes requested):
Hey @mfriesen, firstly thank you for another nice PR. 👍 

Did you change the spacing or the order or something? Everything seems to be changed, while I think it actually is not.

I am working reviewing it, however it would be neat if you could maybe get this PR to be a bit smaller, is that possible? Maybe you could split it up a bit more?

I will add some more comments later on. Thank you.

src/html5sortable.ts
```diff
@@ -577,185 +580,225 @@ export default function sortable (sortableElements, options) {
...
+    _on([sortableElement, placeholder], 'drop', drop)
+
+    _on(items.concat(sortableElement), 'dragover', onDragOverEnter)
+    _on(items.concat(sortableElement), 'dragenter', onDragOverEnter)
+  })
+
+  return sortableElements
+}
+
+function _getHoverClass(sortableElement) {
```
lukasoppermann, line 604: I think the change of the hoverClass should be in a different small PR.
However in this one we could remove the possibility to set options.hoverClass to `true` and always expect a `string`.

This would let you simplify this part a bit.
E.g. the first check could be: 
```
if (options.hoverClass && typeof options.hoverClass === 'string){
  return options.hoverClass
}
```
You could also remove `return null` and check for `undefined`.

src/html5sortable.ts
```diff
@@ -577,185 +580,225 @@ export default function sortable (sortableElements, options) {
...
-      }
+  var options = _data(sortableElement, 'opts')
+  var maxItems = 0
+  // max items
+  if (options.maxItems && typeof options.maxItems === 'number') {
+    maxItems = options.maxItems
+  }
+  if (maxItems && _filter(_getChildren(sortableElement), _data(sortableElement, 'items')).length >= maxItems) {
+    return
+  }
```
lukasoppermann, line 698: I think this could be improved (I know it was like this before).

```js
if (parseInt(options.maxItems) && _filter(_getChildren(sortableElement), _data(sortableElement, 'items')).length >= parseInt(options.maxItems) ) {
   return
}
```

What do you think?

mfriesen:
@lukasoppermann I can definitely break up this pull request by event. IE: dragstart, dragend, drop, hover, onDragOverEnter. I wanted to put it all out there to get your thoughts. if this was the right approach as I'm not a javascript expert.

src/html5sortable.ts
mfriesen: As far as I can tell it still works. That value is only set once. The other way is to use _data(..) to store it. I can do that if you prefer?

src/html5sortable.ts
mfriesen: I want to make sure that
```
element.addEventListener(event, callback)
```
isn't called twice for the same event/callback. It should only skip adding an event if the event already exists.

lukasoppermann:
I did not see anything that was a "wrong" approach from what I can tell. 😃

The whole code base needs to be cleaned up and organized better and I think what you are doing in this PR goes in this direction. 👍 

If you could send this in the smallest pull requests than still make sense, that would be great. This makes it much easier for me to read and test them, and makes merging it much easier. 

Feel free to leave this big one here until everything in here is merged from the smaller PRs (this way my comments also stay).

Looking forward to your PRs and thank you again for the work you put in, that is really neat.

test/events.js
mfriesen: The target is element that was underneath the element being dragged. So you right it should probably be 
```
Object.defineProperty(event, 'target', {value: li, enumerable: true})
```

src/html5sortable.ts
mfriesen: Ok, I'll take a look.

src/html5sortable.ts
mfriesen: You are right, it's double work. placeholders should be replaced with a map, because it allow access to the placeholder inside of any eventhandler. I wasn't familiar enough with the placeholders so I didn't want to break anything. Definitely can be a different PR.

src/html5sortable.ts
mfriesen: You are right.

src/html5sortable.ts
mfriesen: Definitely could be improved. I was trying not to change too much code at once.

mfriesen:
Definitely will. Thanks for creating this project. I've been needing a javascript drag/drop library and this is by far the best one I've found.

PR #323, opened 2018-02-27
changed placeholders to placeholderMap

mfriesen:
Change placeholders to placeholdermap to make lookups easier.
I'm not sure but I also wanted to change
```
_on([sortableElement, placeholder], 'drop', function (e) {
```
to 
```
_on(sortableElement, 'drop', function (e) {
```
It doesn't look like it would affect anything, but I'm not sure.

### DIFF
--- a/src/html5sortable.ts
+++ b/src/html5sortable.ts
@@ -15,7 +15,7 @@ import {makeElement as _html2element, insertBefore as _before, insertAfter as _a
  */
 var dragging
 var draggingHeight
-var placeholders = []
+var placeholderMap = new Map()
 /*
  * remove event handlers from items
  * @param {Array|NodeList} items
@@ -361,7 +361,7 @@ export default function sortable (sortableElements, options) {
     )
 
     _data(sortableElement, 'items', options.items)
-    placeholders.push(placeholder)
+    placeholderMap.set(sortableElement, placeholder)
     if (options.acceptFrom) {
       _data(sortableElement, 'acceptFrom', options.acceptFrom)
     } else if (options.connectWith) {
@@ -407,7 +407,7 @@ export default function sortable (sortableElements, options) {
       // dispatch sortstart event on each element in group
       sortableElement.dispatchEvent(_makeEvent('sortstart', {
         item: dragging,
-        placeholder: placeholder,
+        placeholder: placeholderMap.get(sortableElement),
         startparent: startParent
       }))
     })
@@ -428,7 +428,7 @@ export default function sortable (sortableElements, options) {
       dragging.style.display = dragging.oldDisplay
       delete dragging.oldDisplay
 
-      placeholders.forEach(_detach)
+      Array.from(placeholderMap.values()).forEach(_detach)
       newParent = this.parentElement
 
       if (_listsConnected(newParent, startParent)) {
@@ -458,7 +458,6 @@ export default function sortable (sortableElements, options) {
     // Handle drop event on sortable & placeholder
     // TODO: REMOVE placeholder?????
     _on([sortableElement, placeholder], 'drop', function (e) {
-      var visiblePlaceholder
       if (!_listsConnected(sortableElement, dragging.parentElement)) {
         return
       }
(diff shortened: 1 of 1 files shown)


lukasoppermann:
Hey, @mfriesen this currently has a bug (apart from the styling) because of which the items are moved into other items. If you run `npm run build` and move items you can see that you can "nest" items. I did not yet figure out which line is the issue, but we need to fix this first.

src/html5sortable.ts
```diff
@@ -428,7 +428,7 @@ export default function sortable (sortableElements, options) {
       dragging.style.display = dragging.oldDisplay
       delete dragging.oldDisplay
 
-      placeholders.forEach(_detach)
+      placeholderMap.forEach(_detach)
```
mfriesen: This is wrong.. it should be Array.from(placeholderMap.values()).forEach(_detach);

src/html5sortable.ts
```diff
@@ -512,13 +511,12 @@ export default function sortable (sortableElements, options) {
           _before(element, placeholder)
         }
         // Intentionally violated chaining, it is more complex otherwise
-        placeholders
+        Array.from(placeholderMap.values())
           .filter(function (element) { return element !== placeholder })
           .forEach(_detach)
-      } else {
-        if (placeholders.indexOf(element) === -1 &&
+        if (Array.from(placeholderMap.values()).indexOf(element) === -1 &&
```
mfriesen: woops.. looks like I removed the else accidentally.

src/html5sortable.ts
lukasoppermann: I thought that too, but `placeholderMap.forEach` should call `_detach` with the following argument according to mdn:

- the element value
- the element key
- the Map object being traversed

since the value is the element, it should actually work like this `placeholderMap.forEach(_detach)` right?

mfriesen: I agree it should but it doesn't look like it. Changing to Array.from(placeholderMap.values()).forEach(_detach) works.

src/html5sortable.ts
lukasoppermann: And couldn't you use `placeholderMap.get(sortableElement) !== element`? Would that work, I did not really use Map yet.

mfriesen: I assume you mean for
```
if (Array.from(placeholderMap.values()).indexOf(element) === -1 && !_filter(_getChildren(element), options.items).length)
```
yeah I'm not sure this line confuses me. I'm not sure what is happening there.

src/html5sortable.ts
lukasoppermann: Okay, weird.

src/html5sortable.ts
lukasoppermann: So this part `Array.from(placeholderMap.values()).indexOf(element) === -1` check for element NOT existing in the array.

mfriesen: correct. I have no idea what this does? Length of 0 == true? everything else is false?
```
!_filter(_getChildren(element), options.items).length
```

lukasoppermann: Yes, this is pretty confusing, totally agree. It checks if the list (element) is empty  (= has no child of the type specified as option.items) in it. Because if the list is empty we need to `element.appendChild(placeholder);` Otherwise we insertAfter or Before the hovered item.

This code can probably written in a better and easier to read way, though.

I think having the same function for dragover list and dragover item was not the best idea, this should be changed in the future.

mfriesen:
@lukasoppermann  I've fixed up the branch

src/html5sortable.ts
```diff
@@ -575,9 +574,9 @@ sortable.__testing = {
   _makeGhost: _makeGhost,
   _index: _index,
   _makeEvent: _makeEvent,
-  _getPlaceholders: () => placeholders,
+  _getPlaceholders: () => placeholderMap,
   _resetPlaceholders: () => {
-    placeholders = []
+    placeholderMap = new Map()
```
lukasoppermann, line 579: We could just use `placeholderMap.clear()` here, right?

PR #324, opened 2018-02-27
refactored maxitems

mfriesen:
Refactored maxitems


lukasoppermann:
Thanks, maybe at some point you are interested in looking into #329

Issue #259, opened 2017-05-09
Refactor & review test


lukasoppermann:
- [x] Tests need to be split into smaller, more meaningful files
- [ ] Tests needs to be reviews, what is missing, what is actually tested

jmuzsik:
Hi! I'd be willing to work on this

lukasoppermann:
Hey @jMuzsik, very awesome! I just merged #343 which changed the framework to jest.

I did refactor some tests for units I did extract e.g. `debounce` but the big part is still in the files:

- __tests__/api.test.ts
- __tests__/events.test.ts
- __tests__/internal.test.ts
- __tests__/ghost.test.ts

I think they should all be split up into smaller files which only test specific things. `__tests__/internal.test.ts` should in the long run be completely removed and replaced with tests like the `debounce.test.ts`.

Looking forward to your help. It would be great if you could create a PR immediatly when you push your first branch and just name it with `🛑 [WIP]` in the beginning so that I know it is not ready to be merged. This way I can see what you are working on so that I don't work on the same. 😉 

Also if you can package your changes into small PRs that would be best, as I can merge them faster and easier. 👍 

Just FYI, as I don't know how well you know jest:

- `npm test` runs all files in `__tests__` that are named xyz.test.ts
- `npm run jest` runs jest so you can do `npm run jest something` and it will run `__tests__/something.test.ts`
- `npm run coverage` shows you the test coverage

Thanks mate.

jmuzsik:
Ok! Great, so for example: in Internal function tests:

```
_removeSortableEvents
_removeItemEvents
_removeItemData
...all the rest of the functions
```

all ought to be in their own files, and I will need to slightly modify the code to fit into the Jest framework just as you did with ```debounce```.

Simultaneously, I ought to put more descriptive comments about what is exactly happening within the test? Though, solely if the comments are not already there. (this I cannot guarantee ideal work as I cannot assure that my reading/testing of the code will be the truth).

Or am I reading this wrong: " Tests needs to be reviews, what is missing, what is actually tested
". Is it more nuanced than I thought?

I'm working on several things right now as well, so I'll likely be doing a little every few days if that is alright.

lukasoppermann:
Hey @jMuzsik,

> I'm working on several things right now as well, so I'll likely be doing a little every few days if that is alright.

Sure, I appreciate any help you can put it, no worries.

Adding descriptive comments will definitely be very helpful. Feel free to also improve comments that are already there.

[...]

PR #337, opened 2018-03-02
Fixed Copy:true elements are not sortable after drop

mfriesen:
This Pull Request fixes the #311 issue.  

I changed the drag/drop events to work at the SortableElement level, so when elements are added to a SortableElement, the automatically inherit all events.

### DIFF
--- a/src/html5sortable.ts
+++ b/src/html5sortable.ts
@@ -168,17 +168,19 @@ var _removeItemData = function (items) {
  * @param {Element} destList
  */
 var _listsConnected = function (curList, destList) {
-  var acceptFrom = _data(curList, 'opts').acceptFrom
-  if (acceptFrom !== null) {
-    return acceptFrom !== false && acceptFrom.split(',').filter(function (sel) {
-      return sel.length > 0 && destList.matches(sel)
-    }).length > 0
-  }
-  if (curList === destList) {
-    return true
-  }
-  if (_data(curList, 'connectWith') !== undefined) {
-    return _data(curList, 'connectWith') === _data(destList, 'connectWith')
+  if (_isSortable(curList)) {
+    var acceptFrom = _data(curList, 'opts').acceptFrom
+    if (acceptFrom !== null) {
+      return acceptFrom !== false && acceptFrom.split(',').filter(function (sel) {
+        return sel.length > 0 && destList.matches(sel)
+      }).length > 0
+    }
+    if (curList === destList) {
+      return true
+    }
+    if (_data(curList, 'connectWith') !== undefined) {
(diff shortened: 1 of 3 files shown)


lukasoppermann:
Hey @mfriesen,

thank you very much. It is a nice idea! 
Sadly however it currently breaks the lists with handles (the blue white orange one) as well as the sortable table.

It might be due to using `e.target`, this can be an issue when the target is the `td` or the handle instead of the whole list element. 

One way of solving this could be similar to #338 and adding a `is-sortable-item` attribute to every list item.

mfriesen:
@lukasoppermann Have a look now. I think I got everything working. I added logic to find the correct sortable from the e.target

lukasoppermann:
Nice, will have a look as soon as possible, thank you.

src/html5sortable.ts
```diff
@@ -560,6 +595,8 @@ export default function sortable (sortableElements, options) {
     }, options.debounce)
     // Handle dragover and dragenter events on draggable items
     var onDragOverEnter = function (e) {
+      var element = e.target
```
lukasoppermann, line 591: This element is a `<td>` in tables, which is why it does not work for tables. We need to this because it means tables currently don't work.

Maybe you can just use `.closest(options.item)` because `closest` should return the item, if it matches the selector and go up the dom tree if it does not.

lukasoppermann:
Hey, so I merged #343 which sadly makes this PR "break".

However the adjustments are minor:
1. `_getChildren` was removed, use `.children` instead, e.g. `sortableElement.children`
2. testing framework has been changed to Jest. All tests in the `test` folder have been moved to `__tests__` and "updated" to work with jest. Those are minor updates, so you would just have to add your changes to `test/events.js` to the `__tests__/events.test.ts` file.

Sorry for the inconvenience, I just needed to merge this so #259 can be started. Looking forward to your update. 

Maybe you can just use `.closest(options.item)` or something.

mfriesen:
@lukasoppermann  I've fixed the branch

src/html5sortable.ts
lukasoppermann: This is still a problem, as it still returns `<td>`

lukasoppermann:
Hey, sorry, but I think it is not fixed yet. The table still does not work, as `e.target` is a `<td>` but needs to be a `<tr>`.

mfriesen:
@lukasoppermann I think it's all fixed now.

lukasoppermann:
Neat, thank you. I will give the master branch a bit more manual testing and than we can build another release with the copy feature included. 👍

PR #349, opened 2018-03-05
Corrected spelling error

beyerz:
When copy-pasting from examples, this spelling error causes a break.


lukasoppermann:
Very good catch, thank you. 👍

PR #352, opened 2018-03-06
[WIP] - refactor - modularise tests and improve readability

jmuzsik:
Hi, my commit message is about exactly what it is that I am doing. In this case I did so almost completely with the `internals.test.ts` file. Though, of course, I am still working on this.

Let me know if it is bothersome that I have to repeat code in the `beforeEach` at times. Though, know this is a rough draft as you stated you wanted me to push immediately. I will make sure it looks very clean in the end.

Also, I added `.dist` to `.gitignore`, I can make that is deleted at the end of the request.

And I am not sure while `events.test.ts` is in this request, as I did not write that.

Thanks!


lukasoppermann:
Hey @jMuzsik as I guess you notice, the `events.test.ts` showed up because your branch was out of sync with master. 

While the merge helps, it is better to do a `git rebase master`. This will also potentially turn up some merge conflicts that you can solve the same way, however it will put your changes on top and thus result in a better history. This is why I would ask you to reset to `9d44b54` before the merge and do a git rebase instead.

Also the dist can not be added to gitignore, because on master I do need to push dist for releases.

I will have a more in-depth look, but from what I saw now it looks very good. 👍 
Sadly I have to tell you that the `_attached` function was removed from the code, so the test is unnecessary and will probably fail once you rebased.

Looking forward to you updates, thank you!

jmuzsik:
Ok, I thought I did rebase but, yes, I do not have much experience doing so. I brought my commit back to `9d44b54`. Ok, I'll make sure that the dist and all else is in the right place in the final commit. I'll make sure to delete the `_attached` test.

Thanks for the helpful comments!

lukasoppermann:
Hey, nice. If you need help with the rebase, just let me know.

jmuzsik:
Let me know if I'm on the right track. Hope that I did the right work within the terminal but there is uncertainty.

lukasoppermann:
Hey, sadly the rebase seems not to have worked. You should not have any merges and also not my pushes in your history. (Careful there a renew pushes you need to pull down first).

Do you want to try this again? (Maybe this is good: https://nathanleclaire.com/blog/2014/09/14/dont-be-scared-of-git-rebase/) Otherwise I can have a look and check out how best to rebase it tomorrow.

PR #363, opened 2018-03-10
fixed when element is bigger than draggable and inserting at the top

mfriesen:
fixed when element is bigger than draggable and inserting at the top of sortable.

If you look at the screenshot in the pull request. You'll see if the element is bigger than the draggable. It won't let you insert above it. If you first move the draggable below the element and then move it above the element, it works.

It seems if you remove the deadzone check above the element everything works. I'm not sure of the consequences of this change, but it seems to work better?


lukasoppermann:
Hey, seems to work fine. Once you remove the image I will merge it, thanks.

mfriesen:
@lukasoppermann updated the pull request

Issue #340, opened 2018-03-02
Add touch events


lukasoppermann:
Check & improve how it works on mobile

lukasoppermann:
Duplicate #286

Issue #314, opened 2018-02-23
Convert to typescript


lukasoppermann:
Convert the source file to typescript, which will be compiled in the rollup build step. 
While we currently use `.ts` files, typehinting etc. is not used in the files.

**If you want to help:**
Just start adding typehints to any file that is still missing them or improve the ones that already have them. Feel free to send a PR for files even if only a part of them is converted.

3r3n-n:
Hi ! 
I see that this is an old issue but if there are still some files that need to be converted to TS, I’d be happy to help.

lukasoppermann:
Hey, thank you, there are two important things concerning typescript:
1. The PR to update fails #745 so this needs to be fixed (I didn't really check out what is the problem yet)
2. The main file with everything that was not extracted to smaller modules yet is not using typings or other typescript features at all and could profit from it: https://github.com/lukasoppermann/html5sortable/blob/master/src/html5sortable.ts

If you want to work on either or both of those issues it would be very great.
You can also extract functionality from the html4sortable.ts into smaller modules where it makes sense and convert this over to typescript. This may be easier and more fun than in the big file.

3r3n-n:
Great !
I am an OS newbie so I think I would be more comfortable trying the second option :)

lukasoppermann:
Sounds great. Whatever you feel most comfortable with is a good start and very much appreciated. 👍

If you have questions, just send a PR and add a comment with your questions.

lukasoppermann:
Stale

PR #370, opened 2018-03-13
[WIP] refactor - eslint in all test files, fix some errors

jmuzsik:
Hi!

I mainly integrated additional typescript into the test files with a little refactoring. There is still much to do. 

Let me know if there is something you think I ought to particularly focus on. Otherwise I'll try to do additional refactoring and try to separate tests into folders/new files though it is more complicated with these files than the last one I did.


lukasoppermann:
Hey @jMuzsik, I think you had a small misunderstanding. The `global` in `/* global describe,test,expect */` does not stand for the `global.body` variable but tells eslint that whatever follows are "global" variables and thus standard will not complain about it not being there. So if you re-add it the `CustomEvent` error should vanish. You could also probably use `window.CustomEvent` to get rid of this issue. Standard complains about some other things you would need to fix before I can merge this.

I will look into it and give you some more feedback later. Additional refactoring would definitely be nice.
If you would rather focus on non-test-related issues something like #314 or #340 would be good. (Or any other issue you have a particular interest in). Help with the tests or any of those is very good and appreciated.

Thank you. 😃

lukasoppermann:
Hey @jMuzsik it is still failing. Just in case you did not notice, you can run `npm test` or `npm run standard` and you will see all the issues locally.

jmuzsik:
Ok, I understand, I won't push unless I see that it is working. It is WIP atm, thanks!

lukasoppermann:
Hey @jMuzsik this is not what I meant. Feel free to push in-between versions as much as you want. 😉

jmuzsik:
Ok, got it! I have a difficult time deciphering how to split these tests into different test files as I do not have a very clear idea of what part of the code base each file works in unison with.  So, I would not know how to structure them into different folders nor if I ought to do so.

Also, the code is quite concise already so I do not see much that I can do in terms of modularity. I think I'll work on integrating typescript into other files other then the test files after this. 

Also, I notice that typescript does not agree with the way in which the sortable function is declared seeing how `__testing` is continually a type error within the tests. Reworking that function is also something I can attempt to work with.

Lastly, the type errors that exist other then the ones I fixed are complex. For example, some form of mixin or the like needs to be created to redefine the return value of `new CustomEvent('something')` so that these other properties are expected to occur.

lukasoppermann:
Hey @jMuzsik,

getting more typescript into the code is definitely nice, especially for the non-test files.

The whole package will converted into an es6 module and `__testing` will be removed. So don't worry about this to much.

Currently the most interesting parts would be:
- add types to all functions to catch errors
- clean up the code base, adding comments and refactoring stuff into smaller functions that don't rely on external variables so they can be easily moved around and refactored

jmuzsik:
Ok! I have a slight misunderstanding. What do you mean by catch arrows? Supposing you meant errors?

lukasoppermann:
I should not be doing this at night 😆You are correct, I did mean to write errors.

PR #371, opened 2018-03-13
[WIP] Avoid misfiring 'sortupdate'

mfeherpataky:
That's a first time I work with TypeScript - so no idea how it transcodes into actual JS.

What the commit does - it avoids ```sortupdate``` misfiring by filtering out placeholder item from a sortable list. It brings some changes to filter.ts - where two modes are introduced instead of just calling it a filter and pass a boolean for either finding an intersection or a complement (differences) of a two sets.

I also changed a generic variable ```index``` into something more descriptive ```itemStartIndex```.

Also have doubts about names of detail object keys that is returned on ```sortupdate``` - took me a while to understand what these are. I would suggest:

- ```startIndex```
- ```endIndex```
- ```startSortableIndex``` (only sortable items)
- ```endSortableIndex``` (only sortable items)

Though you probably have some people depending on the current setup, so it would need to be gracefully deprecated.

Anyway - thanks for the code, I am prototyping something for my team your code what the closest one to what I needed! Super helpful. If we stick with it, I would be happy to give you a hand (to the extend I can) in my free time.


lukasoppermann:
Hey, thank you for the PR 👍. However I think it would be more optimal to do it like I outlined below, as it is less change and less added code. Please let me know what you think about it.

Instead of creating a whole getDiff & getIntersect function I would like to just do a `.filter` on the array returned from the `_filter` function, using the actual placeholder from the `placeholderMap` to compare against. The benefits are: We don't need to have two functions and can stick with just `_filter` which make it easier and we don't rely on placeholder having a class (What if the user uses `null` or an empty string as a `placeholderClass`). 

The naming of the properties in the event is awful (I totally agree with you) and I appreciate the change in naming. I am planning to refactor the events and maybe I can improve it even more, but your names are already a definite improvement. I will eventually push this as a minor version update, so this "break" will be okay.

If you find time to help improve and develop this package, that would be a blast 🚀.

mfeherpataky:
Is this what you had in mind?

I also took a liberty and updated the attribute names in the ```detail``` object coming out of ```sortupdate``` event, per the exchange we had.

lukasoppermann:
Hey @mfeherpataky that looks really good. Could you please update the readme with the new names of the items in the event details and remove the compiled files (Js files) from the pull request?

I will do some testing and merge it after those minor updates.

Thanks for the nice work.

mfeherpataky:
@lukasoppermann OK, Done.
I also updated docs/index.html to refer to new values in console.log().

I have one problem with the PR, though, that I could not fix. I did not rename the details.startparent to details.startParent (to follow camelCase naming used elsewhere) for ```sortstart``` event output. Test in ```events.test.ts - should correctly run dragstart event``` was failing no matter how I reassigned the value and just kept on insisting it the value was undefined. Did not have more time to dive into the full test setup - sorry.

So - it works, with that inconsistency where ```sortstart``` returns detail.startparent, and other events return detail.startParent.

--

As for test coverage - there are few more lines in html5sortable.ts, that explains the drop. Though 2% for the whole project is surprising. I did not look at how that number is computed.

lukasoppermann:
Hmm, actually the event is still wrong. When you sort item 1 to position two and afterwards item 1 (from pos 2) to pos 1 again you get 0 as sortableStartIndex and sortableEndIndex. 🤔

PR #375, opened 2018-03-14
[WIP] enhancement - convert source code into typescript

jmuzsik:
This is a beginning attempt in utilising more typescript/getting rid of the red squiggly typescript errors.

Some things that are potentially not ideal:

Added `esnext` into `tsconfig` to get ride of errors that utilise es6 or es2018, etc. Also, I realised that, at times, a custom event can be used to get rid of errors that mention `Event` that does not have a certain method or prototype, etc. 

Thanks! This is a work in progress


lukasoppermann:
Hey, adding esnext makes sense to me! 👍

However I think using customElement and customEvent is not the best. Instead of going generic we should be specific here and use more correct events e.g. DragEvent or define interfaces, e.g. 

I did this for offset in the current PR for setDragImage.

jmuzsik:
Great, though my thoughts were more so that, at times, `Element` type methods are needed so a wrapper could be put around `Element` to still hold the methods while adding new methods to it. 

But it's likely better to make specific interfaces as you said.

jmuzsik:
Ok, I've been having more git problems, and was trying to make sure your commit stayed where it was but it disappeared. Though the code you wrote is still integrated into what I did.

Basically, the main thought I have here is to create an `index.d.ts` file to create the all the types needed. I think this is best because it is one of the only ways in which you can make sure that each `element`s methods and variables are type checked. I believe we can put the entire variety of types within this file so interfaces, etc. will barely have to be used outside of it. I am quite sure this is what is normally done? I know you have one of these files for `jest` in the node modules (`@types/jest`) but that appears to be the only one. Perhaps there are other previously created `.d.ts` files that can be used?

Also, I added more to the `tsconfig`, specifically not allowing `null` or `any`. I can delete that afterwards. This is a beginning. But I believe that I am working with too many files here. For this PR I think it's best if I solely work on the `.d.ts` file and get that up and running. Unless you see a better way to go about this?

lukasoppermann:
Hey @jMuzsik if you rebase correctly, my commit **should** disappear from your PR. You should only see your commits. 

**What `rebase master` does:**
- "remove" your commits from the branch
- reset the branch to the current version of the other branch e.g. `master`
- add your commits on top of it

So it basically is like you would have branched out just now.

Creating a separate file with typescript interfaces makes sense. However, I think it would be better to create a folder with all interfaces in separate files per interface and import all of those into `index.d.ts`, which makes it easier for us. 

`@types/jest` is an external resource that we don't need to touch or integrate further.

I think not allowing `null` or `any` will not work as we need this for some cases. However disabling "assumed any" (as in, we did not put anything so typescript puts in any) could be done in the future.

Focusing this PR on adding interfaces is a good idea. Maybe you could even start with just the `index.d.ts` and one interface in a subfolder. I will be able to merge this quickly and you can send more of those in other PRs (small PRs are much faster and easier to check & merge).
Also we should include it in the `package.json` as it is recommended: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

Also maybe you could do a first PR which adds typescript to the npm scripts, so that `npm run ts` does a sensible typescript check (of everything in src) but is not used in travis, so that it does not fails builds. We can only add it to travis once we have everything working.

Thanks for your effort, looking forward to your PRs. 👍

lukasoppermann:
Can you please create a small intermediate PR with the following so I can easily merge it:

1. Your `tsconfig.json` changes (lets go ahead with `strictNullChecks`) as well. We can still remove it if we need to later on
2. Your fix for `src/makePlaceholder.ts`
3. Your improvement to `src/index.ts`
4. `filter.ts` if you:
- remove the comment on the last line
- change `@returns {Node[]|Element[]}` to `@returns {Array<Element>}` (they are all elements, because we filter by nodeType === 1)

If you send this in one separate PR I can quickly merge it, thanks. 👍

jmuzsik:
Hi, I realised that I did some additional work since the last time I committed, I can delete it prior, just let me know. 

Thanks for all of the explanations, really appreciate it!

I'll actually make a separate PR and this one can be scraped, that may be best.

PR #383, opened 2018-03-19
Addition of npm script for tsc checking

jmuzsik:
`tsc ./src/*.ts --pretty --diagnostics --noEmit --watch`

screen shot 2018-03-19 at 2 49 59 pm
screen shot 2018-03-19 at 2 49 45 pm

`--pretty` does just as it says, makes the output look better in the terminal.

`--diagnostics` is simply that nice little note at the end of the output.

`--noEmit` prevents the creation of .js files for each ts compilation.

And that's all, lemme know if you want me to add/delete something. I basically went through [this](https://www.typescriptlang.org/docs/handbook/compiler-options.html) file and added what I thought was relevant.

Also, I ran `--declarationDir` to get an output that created about exactly what you were talking about, an individual `.d.ts` file for each ts file in a new directory, as a good starting point.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "cover": "jest --colors --coverage --coverageReporters=text-lcov | coveralls",
     "preversion": "npm test",
     "version": "npm run build && git add -A dist docs",
-    "postversion": "git push && git push --tags"
+    "postversion": "git push && git push --tags",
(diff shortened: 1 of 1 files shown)


lukasoppermann:
Thanks, perfect. 👍

Issue #326, opened 2018-02-27
Convert package into ES6 Class


lukasoppermann:
This will go in line with the new api #345 

The idea is to turn sortable into an ES6 class which will not only give us some helpful features, but also make the whole package more structured and easier to read and understand.

PR #390, opened 2018-03-20
beginning of conversion of src files to ts, utilising interfaces

jmuzsik:
Hi, this is a more informed TS effort!

I added `super` in `setDragImage` test because my tests continually failed otherwise while ESNext was on.

Otherwise, lots of type checks to get rid of most of the warnings that existed in each file. Most files do not have warnings atm.

Though, this is going to bring down test coverage as there are quite a few lines added without tests. I can add tests as well as soon as I can to help get that coverage up!

Also, the `tsconfig` has the part with `src/interfaces` as, I am quite sure, that is necessary to keep reference to `.d.ts` files.

Thanks!

### DIFF
--- a/src/attribute.ts
+++ b/src/attribute.ts
@@ -1,9 +1,9 @@
 /**
- * @param {Array|elem} element
+ * @param {Array<HTMLElement>|HTMLElement} element
  * @param {string} attribute
- * @param {*} value
+ * @param {string} value
  */
-function addAttribute (element: Array<Element>|Element, attribute:string, value:string) {
+function addAttribute (element: Array<HTMLElement>|HTMLElement, attribute:string, value:string) {
   if (element instanceof Array) {
     for (var i = 0; i < element.length; ++i) {
       addAttribute(element[i], attribute, value)
@@ -13,10 +13,10 @@ function addAttribute (element: Array<Element>|Element, attribute:string, value:
   element.setAttribute(attribute, value)
 }
 /**
- * @param {Array|Element} element
+ * @param {Array|HTMLElement} element
  * @param {string} attribute
  */
-function removeAttribute (element: Array<Element>|Element, attribute:string) {
+function removeAttribute (element: Array<HTMLElement>|HTMLElement, attribute:string) {
   if (element instanceof Array) {
     for (var i = 0; i < element.length; ++i) {
       removeAttribute(element[i], attribute)
--- a/src/attribute.ts
+++ b/src/attribute.ts
@@ -1,9 +1,9 @@
 /**
- * @param {Array|Element} element
+ * @param {Array|elem} element
  * @param {string} attribute
  * @param {*} value
  */
-function addAttribute (element, attribute, value) {
+function addAttribute (element: Array<Element>|Element, attribute:string, value:string) {
   if (element instanceof Array) {
     for (var i = 0; i < element.length; ++i) {
       addAttribute(element[i], attribute, value)
@@ -16,7 +16,7 @@ function addAttribute (element, attribute, value) {
  * @param {Array|Element} element
  * @param {string} attribute
  */
-function removeAttribute (element, attribute) {
+function removeAttribute (element: Array<Element>|Element, attribute:string) {
   if (element instanceof Array) {
     for (var i = 0; i < element.length; ++i) {
       removeAttribute(element[i], attribute)
--- a/src/data.ts
+++ b/src/data.ts
@@ -6,7 +6,7 @@
  * @return {*}
  */
 
-function addData (element: HTMLElement, key: string, value?: any): HTMLElement|opts|string|void {
+function addData (element: HTMLElement, key: string, value?: any): HTMLElement|configuration|string|void {
   if (value === undefined) {
     return element && element.h5s && element.h5s.data && element.h5s.data[key]
   } else {
--- a/src/data.ts
+++ b/src/data.ts
@@ -1,12 +1,12 @@
 /**
  * Get or set data on element
- * @param {Element} element
+ * @param {HTMLElement} element
  * @param {string} key
- * @param {string} value
+ * @param {any} value
  * @return {*}
  */
 
-function addData (element: Element, key: string, value?: string): Element|opts|string|void {
+function addData (element: HTMLElement, key: string, value?: any): HTMLElement|opts|string|void {
   if (value === undefined) {
     return element && element.h5s && element.h5s.data && element.h5s.data[key]
   } else {
@@ -17,9 +17,9 @@ function addData (element: Element, key: string, value?: string): Element|opts|s
 }
 /**
  * Remove data from element
- * @param {Element} element
+ * @param {HTMLElement} element
  */
-function removeData (element: Element) {
+function removeData (element: HTMLElement) {
   if (element.h5s) {
     delete element.h5s.data
   }
--- a/src/data.ts
+++ b/src/data.ts
@@ -2,10 +2,11 @@
  * Get or set data on element
  * @param {Element} element
  * @param {string} key
- * @param {*} value
+ * @param {string} value
  * @return {*}
  */
-function addData (element: Element, key: string, value?: any) {
+
+function addData (element: Element, key: string, value?: string): Element|opts|string|void {
   if (value === undefined) {
     return element && element.h5s && element.h5s.data && element.h5s.data[key]
   } else {
--- a/src/debounce.ts
+++ b/src/debounce.ts
@@ -1,13 +1,12 @@
 /**
  * Creates and returns a new debounced version of the passed function which will postpone its execution until after wait milliseconds have elapsed
- * @param {func} Function to debounce
+ * @param {Function} func to debounce
  * @param {number} time to wait before calling function with latest arguments, 0 - no debounce
  * @returns {function} - debounced function
  */
 export default (func: Function, wait: number = 0): Function => {
-  let timeout = wait
-
-  return (...args: Array<Element>) => {
+  let timeout
+  return (...args) => {
     clearTimeout(timeout)
     timeout = setTimeout(() => {
       func(...args)
--- a/src/debounce.ts
+++ b/src/debounce.ts
@@ -1,13 +1,13 @@
 /**
  * Creates and returns a new debounced version of the passed function which will postpone its execution until after wait milliseconds have elapsed
  * @param {func} Function to debounce
- * @param {wait} time to wait before calling function with latest arguments, 0 - no debounce
+ * @param {number} time to wait before calling function with latest arguments, 0 - no debounce
  * @returns {function} - debounced function
  */
 export default (func: Function, wait: number = 0): Function => {
-  let timeout
+  let timeout = wait
 
-  return (...args) => {
+  return (...args: Array<Element>) => {
     clearTimeout(timeout)
     timeout = setTimeout(() => {
       func(...args)
--- a/src/elementHeight.ts
+++ b/src/elementHeight.ts
@@ -3,7 +3,7 @@
  * Get height of an element including padding
  * @param {Element} element an dom element
  */
-export default element => {
+export default (element: Element) => {
   if (!(element instanceof Element)) {
     throw new Error('You must provide a valid dom element')
   }
--- a/src/elementHeight.ts
+++ b/src/elementHeight.ts
@@ -1,10 +1,10 @@
 /* eslint-env browser */
 /**
  * Get height of an element including padding
- * @param {Element} element an dom element
+ * @param {HTMLElement} element an dom element
  */
-export default (element: Element) => {
-  if (!(element instanceof Element)) {
+export default (element: HTMLElement) => {
+  if (!(element instanceof HTMLElement)) {
     throw new Error('You must provide a valid dom element')
   }
   // get calculated style of element
--- a/src/eventListener.ts
+++ b/src/eventListener.ts
@@ -1,8 +1,8 @@
 import store from './store'
 /**
  * @param {Array|HTMLElement} element
- * @param {string} event
  * @param {Function} callback
+ * @param {string} event
  */
 
 function addEventListener (element: Array<HTMLElement>|HTMLElement, event:string, callback: () => void) {
--- a/src/eventListener.ts
+++ b/src/eventListener.ts
@@ -1,11 +1,11 @@
 import store from './store'
 /**
- * @param {Array|Element} element
+ * @param {Array|HTMLElement} element
  * @param {string} event
  * @param {Function} callback
  */
 
-function addEventListener (element: Array<Element>|Element, event:string, callback: () => void) {
+function addEventListener (element: Array<HTMLElement>|HTMLElement, event:string, callback: () => void) {
   if (element instanceof Array) {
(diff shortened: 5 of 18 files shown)


src/attribute.ts
```diff
@@ -1,9 +1,9 @@
 /**
- * @param {Array|Element} element
+ * @param {Array|elem} element
```
lukasoppermann: Why are you changing it to elem? Shouldn't you rather change it to `Array<HTMLElement>|HTMLElement`?

src/attribute.ts
```diff
@@ -1,9 +1,9 @@
...
  * @param {string} attribute
  * @param {*} value
```
lukasoppermann: this should be `string` instead of `*` right?

src/attribute.ts
```diff
@@ -1,9 +1,9 @@
...
  */
-function addAttribute (element, attribute, value) {
+function addAttribute (element: Array<Element>|Element, attribute:string, value:string) {
```
lukasoppermann: I think it should be `element: Array<HTMLElement>|HTMLElement`.

src/attribute.ts
```diff
@@ -16,7 +16,7 @@ function addAttribute (element, attribute, value) {
  * @param {Array|Element} element
  * @param {string} attribute
  */
-function removeAttribute (element, attribute) {
+function removeAttribute (element: Array<Element>|Element, attribute:string) {
```
lukasoppermann: Same as above: `Array<HTMLElement>|HTMLElement`

src/data.ts
```diff
@@ -2,10 +2,11 @@
  * Get or set data on element
  * @param {Element} element
  * @param {string} key
- * @param {*} value
+ * @param {string} value
```
lukasoppermann: This needs to be `any`

src/data.ts
```diff
@@ -2,10 +2,11 @@
  * Get or set data on element
  * @param {Element} element
```
lukasoppermann: We should be using `HTMLElement`

src/data.ts
```diff
@@ -2,10 +2,11 @@
...
  * @param {string} key
- * @param {*} value
+ * @param {string} value
  * @return {*}
  */
-function addData (element: Element, key: string, value?: any) {
+
+function addData (element: Element, key: string, value?: string): Element|opts|string|void {
```
lukasoppermann: Same here: `HTMLElement` and value must be `any`

src/elementHeight.ts
```diff
@@ -3,7 +3,7 @@
  * Get height of an element including padding
  * @param {Element} element an dom element
  */
-export default element => {
+export default (element: Element) => {
```
lukasoppermann: Again `HTMLElement` 😉

src/debounce.ts
```diff
@@ -1,13 +1,13 @@
 /**
  * Creates and returns a new debounced version of the passed function which will postpone its execution until after wait milliseconds have elapsed
  * @param {func} Function to debounce
- * @param {wait} time to wait before calling function with latest arguments, 0 - no debounce
+ * @param {number} time to wait before calling function with latest arguments, 0 - no debounce
  * @returns {function} - debounced function
  */
 export default (func: Function, wait: number = 0): Function => {
-  let timeout
+  let timeout = wait
```
lukasoppermann: this should not be wait

src/debounce.ts
```diff
@@ -1,13 +1,13 @@
...
 
-  return (...args) => {
+  return (...args: Array<Element>) => {
```
lukasoppermann: We can say what the arguments are, so we need to remove the `Array<Element>`

src/eventListener.ts
```diff
@@ -19,7 +20,7 @@ function addEventListener (element, event, callback) {
  * @param {Array|Element} element
  * @param {Array|string} event
  */
-function removeEventListener (element, event) {
+function removeEventListener (element: Element, event: string) {
```
lukasoppermann: Element should be `Array<HTMLElement>|HTMLElement`

src/index.ts
```diff
@@ -4,7 +4,7 @@
  * @param {Element} element
  * @returns {number}
  */
-export default (element: Element, elementList: HTMLCollection | NodeList | Array): number => {
+export default (element: Element, elementList: HTMLCollection | NodeList | Array<Element>): number => {
```
lukasoppermann: `Array<HTMLElement>`

lukasoppermann (changes requested):
Hey, thank you for all the work!
I will look into the other files as well. 

I would prefer to use `types` instead if `interfaces` as a folder name, like in this PR: https://github.com/lukasoppermann/html5sortable/pull/387/files#diff-86b55afca5305c6cfb991f9b13264803

From my research `types` is a more typical folder for this. Please let me know if you have concerns.

Also, I thought we figured it would be best to create individual files for each interface (similar to the one I have in this PR, although I don't have the `declare` part which I will look into 👍). Would you mind updating or is there a reason for this?

Can you please rebase, I did some changes, that will influence your code in `makePlaceholder.ts` & `insertHtmlElements.ts`.

src/filter.ts
```diff
@@ -5,13 +5,14 @@
  * @param {String} selector
  * @returns {Array}
  */
-export default (nodes: NodeList|HTMLCollection|Array, selector: string): Array<Element> => {
+export default (nodes: NodeList|HTMLCollection|Array<Element>, selector: string): Array<Element> => {
```
lukasoppermann: This should be `Array<HTMLElement>`

src/filter.ts
```diff
@@ -5,13 +5,14 @@
...
   if (!(nodes instanceof NodeList || nodes instanceof HTMLCollection || nodes instanceof Array)) {
     throw new Error('You must provide a nodeList/HTMLCollection/Array of elements to be filtered.')
   }
+  let elArray:Element[] = Array.from(nodes)
```
lukasoppermann: Does this actually solve type issues? It does not for me.

src/getHandles.ts
```diff
@@ -4,7 +4,7 @@
  * @param {NodeList|HTMLCollection|Arra} items
  * @param {string} selector
  */
-export default (items: NodeList|HTMLCollection|Array, selector: string): Array<Element> => {
+export default (items: NodeList|HTMLCollection|Array<Element>, selector: string): Array<Element> => {
```
lukasoppermann: I updated this file, I think.

src/insertHtmlElements.ts
```diff
@@ -5,7 +5,7 @@
  * @param {Element} newElement - element to be inserted
  * @param {String} position - insert before or after reference element
  */
-let insertNode = (referenceNode: any, newElement: any, position: String) => {
+let insertNode = (referenceNode: Element, newElement: Element, position: String) => {
```
lukasoppermann: I updated this file as well already, sorry.

src/offset.ts
```diff
@@ -1,9 +1,4 @@
-export interface offsetObject { // eslint-disable-line no-undef
-  'left': number,
-  'right': number,
-  'top': number,
-  'bottom': number
-}
+// eslint-disable-line no-undef
```
lukasoppermann: This comment is not needed if you remove the interface.

src/setDragImage.ts
```diff
@@ -1,5 +1,5 @@
 /* eslint-env browser */
-import {default as offset, offsetObject} from './offset' // eslint-disable-line no-unused-vars
+import {default as offset} from './offset' // eslint-disable-line no-unused-vars
```
lukasoppermann: I think the comment `// eslint-disable-line no-unused-vars` can be removed if you remove `offsetObject `

lukasoppermann (changes requested):

jmuzsik:
Ok! Think this is a much better implementation then the previous one! Open to any more changes.

`Element` changed everywhere to `HTMLElement, maybe I shouldn't have done so in the `html5sortable` file, let me know. 

`Type` folder created. For the few interfaces currently that are used only in specific files I created unique files.

Also, you know that `.d.ts` files can automatically be created by running: `tsc --declaration --outDir ./src/{something}`
screen shot 2018-03-21 at 12 18 26 pm

I made a `main.d.ts` as extensions to objects that are externally created such as `HTMLElement` can be done as so. Also, since it is used in multiple files I thought it best to do the defining in one file. Perhaps it is my editor but if I define an interface one time it does not need to be imported. I do not think the change in the `tsconfig` makes any difference in this case actually as I tried with and without it.

Also, lemme know if you care for the type guard. I can delete it, and I do not think it ought to belong in the `types` folder if it stays put:[ type guards](https://stackoverflow.com/questions/32713452/user-defined-type-guards-typescript)

Thanks!

lukasoppermann (commented):
Hey, you will definitely need to git rebase, my changes should not be in your PR. I added some more minor comments, will check further tomorrow. Looking good already. 👍

src/eventListener.ts
```diff
@@ -16,10 +17,10 @@ function addEventListener (element, event, callback) {
   element.h5s.events[event] = callback
 }
 /**
- * @param {Array|Element} element
- * @param {Array|string} event
+ * @param {Array<HTMLElement>|HTMLElement} element
+ * @param {Array<HTMLElement>|string} event
  */
-function removeEventListener (element, event) {
+function removeEventListener (element: HTMLElement, event: string) {
```
lukasoppermann, line 22: This should also be `element: Array<HTMLElement>|HTMLElement` I think.

src/filter.ts
```diff
@@ -5,13 +5,14 @@
  * @param {String} selector
  * @returns {Array}
  */
-export default (nodes: NodeList|HTMLCollection|Array, selector: string): Array<Element> => {
+export default (nodes: NodeList|HTMLCollection|Array<HTMLElement>, selector: string): Array<HTMLElement> => {
   if (!(nodes instanceof NodeList || nodes instanceof HTMLCollection || nodes instanceof Array)) {
     throw new Error('You must provide a nodeList/HTMLCollection/Array of elements to be filtered.')
   }
+  let elArray:Element[] = Array.from(nodes)
```
lukasoppermann: Sorry to be picky, but can you rename `elArray` to just `elements` I think it is more consistent and makes it easier to understand.

src/html5sortable.ts
```diff
@@ -39,7 +39,7 @@ var _removeItemEvents = function (items) {
 }
 /**
  * Remove event handlers from sortable
- * @param {Element} sortable a single sortable
+ * @param {HTMLElement} sortable a single sortable
  */
 var _removeSortableEvents = function (sortable) {
```
lukasoppermann: We could add a type here: `function (sortable: HTMLElement) {`

jmuzsik:
Think I finally understand interactive rebase, that's mainly thanks to you, thanks! And I did the quick code fix on github by spontaneously, probably should have did it in my own editor.

src/eventListener.ts
```diff
@@ -1,9 +1,10 @@
 /**
- * @param {Array|Element} element
- * @param {Array|string} event
+ * @param {Array|HTMLElement} element
+ * @param {string} event
  * @param {Function} callback
  */
-function addEventListener (element, event, callback) {
+
+function addEventListener (element: Array<HTMLElement>|HTMLElement, event:string, callback: () => void) {
```
lukasoppermann: Are you sure the callback part is correct? Can you explain it? Shouldn't it be callback: Function?

I guess you look this up somewhere so I probably just don't know about it, maybe you can point me to the resource?

src/serialize.ts
```diff
@@ -19,9 +20,15 @@ export default (sortableContainer: Element, customItemSerializer: Function = (se
   }
   // get options
   let options = _data(sortableContainer, 'opts')
+
```
lukasoppermann, line 22: This type guard is not nessesary, as the options part will be replaced with the store.

tsconfig.json
```diff
@@ -1,5 +1,11 @@
 {
   "compilerOptions": {
-    "module": "commonjs"
-  }
+    "module": "commonjs",
+    "target": "ESNext",
+    "strictNullChecks": true
+  },
+  "typeRoots": [
+    "./src/interfaces",
```
lukasoppermann: This should be types now. 😉

lukasoppermann (commented):
Just some minor last things and we can merge it, very nice work.

src/types/main.d.ts
```diff
@@ -0,0 +1,59 @@
...
+    hoverClass: boolean,
+    debounce: number,
+    maxItems: number,
+    itemSerializer: void,
+    containerSerializer: void,
+    items: string
+}
+
+// Can these functions be better defined?
+interface events {
```
lukasoppermann: I think this is not needed, as we can use `DragEvent` which should have those functions, right?

src/types/main.d.ts
```diff
@@ -0,0 +1,59 @@
...
+ * typescriptlang.org/docs/handbook/declaration-merging.html
+ * ie. if this is done new properties can be 
+ * added to previously defined objects
+*/
+interface HTMLElement {
+    isSortable: boolean,
+    h5s: h5s
+}
+
+interface Node {
```
lukasoppermann: Node does not have `querySelector`, so we need to remove this.

src/types/main.d.ts
```diff
@@ -0,0 +1,59 @@
...
+interface h5s {
+    events: events,
+    data: data
+}
+/* 
+ * typescriptlang.org/docs/handbook/declaration-merging.html
+ * ie. if this is done new properties can be 
+ * added to previously defined objects
+*/
+interface HTMLElement {
```
lukasoppermann: I think we should not extend `HTMLElement`. Rather, in the future we should have an interface for the sortable class #326

src/types/main.d.ts
```diff
@@ -0,0 +1,59 @@
+/*
+ * For default interfaces that are frequently used and extensions:
+ *   - HTMLElement: h5s and its inner objects.
+ *   - Extension of objects that are external to the library.
+*/
+
+interface opts {
```
lukasoppermann: could you rename this one to `configuration`?  I know we currently use otps but I would like to change this as well. Thanks

src/eventListener.ts
jmuzsik: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html

And it says to put parameters as well if you know what the callback will be, but it appears that the function is not always the same.

src/serialize.ts
jmuzsik: ok deleted

tsconfig.json
jmuzsik: Maybe best to delete this b/c it does not appear to affect my linter.

src/types/main.d.ts
jmuzsik: ok deleted

src/types/main.d.ts
jmuzsik: I can look into that next

lukasoppermann (approved):

lukasoppermann:
Looks very good, thank you for all the work @jMuzsik 👍

Issue #381, opened 2018-03-15
Sortupdate event index wrong


lukasoppermann:
The values returned in the sortupdate are not always correct.

mfeherpataky:
The problem is with startSortableIndex, right?
I will look at it in the next few days.

lukasoppermann:
Exactly. Thanks!

By the way, I would like to move all the event functions outside of the sortable so that they are only accessing things that are passed into them. If you while working on this issue want to work towards that, that would be neat.

mfeherpataky:
warning - long :grey_exclamation:

While there is a simple fix - just a simple refresh of sortable items cache (var items) on dragstart.
It only works for items moved within a single sortable list. The `sortupdate` is fired off on `dragend` or `drop` in a scope of a sortable item and depends on some local cache-like variables which are set on `dragstart`.

As you you are suggesting - a proper fix probably requires to gut the event function completely from any one sortable. In some cases the event function would need to get values from multiple sortables. 

The cases I see:

a) Item moved within the same list (data from a single sortable list)
b) Item moved from another sortable list where we would want to track both lists (need to pass data from both sortable lists)
c) Item moved form another sortable list where items are copied from, so no need to track both lists. (single list?)

---

The detail object has the following:
- `item` - the element we moved around
- `startParent` - starting List (works as it is held outside of a sortable scope)
- `endParent` - target List
-  :warning: `startIndex` - starting Index within the endParent list (that is set on dragstart in the endParent sortable, which is not fired when dropping in the endParent)
- `endIndex` - target Index within the endParent list
- :warning: `startSortableIndex` - starting Index in the endParent list (that is set on dragstart in the endParent sortable, which is not fired when dropping in the endParent) 
- `endSortableIndex` - target Index in the endParent list

What do you expect to get with startIndex and startSortableIndex in case (b) & (c) - when item is moved across lists? I would expect indexes in the startParent lists, which is not the case.

Going one step forward - there are three serialised lists:
- :warning: `oldStartList` - currently points to the pre-update state of endParent list... Not what I think is expected. May also be undefined as `dragstart` was not fired in the endParent
- `newStartList` - points to post-update state of the startParent list (that's OK)
- `newEndList` - points to post-update state of endParent list (that's OK)

----

**TL;DR I would suggest:**
- a small refactoring of local variables maybe with some further renaming to clarify scope - setting them at the top of the sortable with ES6 let styles and some comments,
[...]

PR #391, opened 2018-03-20
[WIP] Bug/381 sortupdate event index wrong

### LINKED ISSUE #281: oldindex retains original index after sortupdate with multiple moves.
Greetings,

e.detail.oldindex seems to retain the original index of the item that was moved if an item is moved more than once within a list.  Note that e.detail.oldElementIndex does appear to be updated with each move.  

I am working with items in a single list.  Observed this behavior in Firefox and Chrome, have not tried others.

Not sure if this is by design as I could see it being useful referring back to what the lists original order was but it does not sound like it from the documentation which simply states that it contains the old index of the dragged element (considering only list items).

If people do want the original order retained, perhaps an originalIndex property?

mfeherpataky:
Following conversation in #381, also fixes #281.

Tested with linked lists, also with copy-only lists.
Not sure if all the cases were covered, so it needs some more manual tests.

--

FYI - Automatic tests are failing for ghost handling. 
Not sure if this was after a rebase on the new master, but I suspect it was as no ghost code was altered.

Unfortunately for the next few days will not have any free time to continue.

TL;DR - needs manual tests and if works as expected - update to documentation & automated tests code

### DIFF
--- a/src/html5sortable.ts
+++ b/src/html5sortable.ts
@@ -398,7 +398,10 @@ export default function sortable (sortableElements, options: object|string|undef
       const visiblePlaceholder = Array.from(stores.values()).map(data => data.placeholder)
         .filter(placeholder => placeholder instanceof HTMLElement)
         .filter(isInDom)[0]
-      visiblePlaceholder.remove()
+
+      if (visiblePlaceholder) {
+        visiblePlaceholder.remove()
+      }
 
       // dispatch sortstart event on each element in group
       sortableElement.dispatchEvent(new CustomEvent('sortstop', {
--- a/src/html5sortable.ts
+++ b/src/html5sortable.ts
@@ -50,15 +50,6 @@ const _removeItemEvents = function (items) {
   _off(items, 'mouseleave')
 }
 /**
- * Remove event handlers from sortable
- * @param {Element} sortable a single sortable
- */
-const _removeSortableEvents = function (sortable) {
-  _off(sortable, 'dragover')
-  _off(sortable, 'dragenter')
-  _off(sortable, 'drop')
-}
-/**
  * _getDragging returns the current element to drag or
  * a copy of the element.
  * Is Copy Active for sortable
--- a/src/html5sortable.ts
+++ b/src/html5sortable.ts
@@ -395,7 +395,7 @@ export default function sortable (sortableElements, options: object|string|undef
       dragging.style.display = dragging.oldDisplay
       delete dragging.oldDisplay
 
-      const visiblePlaceholder = Array.from(stores.values()).map( data => data.placeholder )
+      const visiblePlaceholder = Array.from(stores.values()).map(data => data.placeholder)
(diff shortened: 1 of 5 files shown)


lukasoppermann:
I did not have time to test it yet, but having a look at the code it looks very good. 👍 
I am currently using `sortableContainer` for what you refer to as `sortableList` in the serialize function. I think Container might be better, because you can technically sort tables and grids as well. What do you think?

I think we need to add some development notes in some files and and link those in the readme, which should include naming, etc. I will start on this but if you have ideas for what you think would make working on it easier, please let me know. For e.g. concerning the setup and typescript.

mfeherpataky:
Yeap, I will refer to `sortableContainers` - that makes sense, especially as that's how we are naming them in the event responses.

I will come back to in on the weekend.
Did you find anything else I should fix/look into for this PR?

edit: Also - I will use `dragenter`, instead of `dragover` to fetch `itemsBeforeUpdate` for destinationContainer. The former is triggered only once and the latter every couple 100ms, with the same result. Just read specs on MDN 😄.

src/html5sortable.ts
```diff
@@ -324,37 +339,58 @@ export default function sortable (sortableElements, options: object|string|undef
...
+          item: dragging
         }
       }))
     })
-    // Handle drag events on draggable items
+
+    /*
+     We are capturing targetSortable data with 'dragover' event
+    */
+    _on(sortableElement, 'dragover', (e) => {
```
mfeherpataky: That will be changed to dragenter - much cheaper to run

lukasoppermann:
Hey, good catch! 👍 
The only thing apart from the `dragenter`, that I can think of, is updating the `readme` and maybe improving the examples so that testing this is easier.

I did check it out and it worked quite well in my test, very neat.

mfeherpataky:
hi @lukasoppermann 

Things done:
1. Rebased onto the latest master,
2. Renamed local variables to sortableContainer - from sortableList
3. Using `dragenter` to fire off less events when getting targetContainer itemsBeforeUpdate
4. Fixed an issue, where in the originContainer the itemsBeforeUpdate still carried placeholder
5. Updated README and docs/index.html to use new event object in console.log

--

However, I am a bit at a loss now with the tests results.
There are some local errors I get (around Ghost and Dragging) that never show up in Travis and vice-versa (export/import issues). To be honest, I cannot figure the tests out and could not since the beginning.

EDIT: Tests are passing now, though the early error messages in Travis had little - if anything - to do with the code I changed 😕.

mfeherpataky:
I found still one issue with this branch - if `dragend` is called second (in cases of a successful drop) then it still attempts to remove the placeholder. Since it is not checking if the placeholder is still there, console errors out on calling .remove() on undefined.

Will fix it and push again

lukasoppermann:
Very nice, thank you for the great work on this. 😃 This improves the events quite a lot!

Issue #384, opened 2018-03-19
Can IE Fix be removed?


lukasoppermann:
Can we remove this IE fix?
 https://github.com/lukasoppermann/html5sortable/blob/master/src/html5sortable.ts#L170-L186

If not, why is it important? Can we explain in code comments why this is needed and for up to which IE versions?

CWSites:
What was the issue that caused you to add it in the first place?

lukasoppermann:
Hey @CWSites, there were some issues with the ghost / dragImage in IE, I don't recall exactly what though. I guess it is not needed in modern IE and Edge browsers anymore.

kaffarell:
Issue closed because it is not present anymore in recent browsers

Issue #351, opened 2018-03-06
Add drag cursor on hover?


lukasoppermann:
By default a cursor should indicate item that an item is draggable and change on drag. 

This turns out to be very tricky. The states that need to be supported:

**Must:**
- [ ] draggble: `cursor: -moz-grab; cursor: -webkit-grab; cursor: grab;`
- [ ] dragged `cursor: -moz-grabbing; cursor: -webkit-grabing; cursor: grabing;`

**Nice to have:**
- [ ] copy: when item is moved into list where it can be copied, show copy cursor
- [ ] not-allowed: show not allowed cursor when moving somewhere on a sortable where the item can not be dropped.

**Maybe this helps:**
- https://stackoverflow.com/questions/44447210/custom-cursor-with-drag-and-drop-an-html-element-without-libraries

CWSites:
I was able to accomplish this by adding the following to my CSS

```
.is-hovered {
    cursor: grab;
    cursor: -moz-grab;
    cursor: -webkit-grab;
}
.sortable-placeholder {
    cursor: grabbing !important;
    cursor: -moz-grabbing !important;
    cursor: -webkit-grabbing !important;
}
```

lukasoppermann:
Can you upload an example of this? I remember having some issues when dragging.

CWSites:
My config was as follows. I'll setup a codepen with a sample.

```
    sortable('.sortable', {
      forcePlaceholderSize: true,
      placeholderClass: 'sortable-placeholder',
      hoverClass: 'is-hovered'
    });
```

CWSites:
Here is a codepen: https://codepen.io/CWSites/full/wjeNbb/ 

I noticed some odd behavior though, the sortable area is only the top row of objects even though I have adjusted the styles to be full width. The dark gray background is the sortable container but notice the red border only appears at the top. Similarly the **grabbing cursor** style only works in that top box and not 100% of the time.

CWSites:
Hey @lukasoppermann I fixed all of the issues and improved the sortable-placeholder as well. Check out my referenced codepen above for a working example.

Configuration
```
sortable('.sortable', {
    forcePlaceholderSize: true,
    placeholderClass: 'sortable-placeholder',
    hoverClass: 'is-hovered'
});
```

Styles Applied
```
// Class of the element that is sortable
.box {
    cursor: grabbing;
    cursor: -moz-grabbing;
    cursor: -webkit-grabbing;
    &.is-hovered {
[...]

Issue #341, opened 2018-03-02
Make clicking work


lukasoppermann:
Clicking on links/buttons should work, while they are still draggable.

This could potentially work by only firing sort start even if the user has a mousedown event and moves the mouse x pixels.

Issue #313, opened 2018-02-23
Revert on invalid target


lukasoppermann:
When an item is dropped outside of any valid sortable it should be moved to its original position (not its last valid position). This should maybe be done using an option.

```js
opts.invalidDrop = 'revert'
```

Other options for `invalidDrop` could be `false` (default which uses last valid position).

And as an extension `remove` which removes the item. (This should probably be a separate PR).

PR #397, opened 2018-03-25
[WIP] Sortable class and conversion to ES6

jmuzsik:
Hi! 

The primary focus is to convert the `sortable` function into a class. Basically,  I aimed for smaller modules at first then settled for big ones in the end that I can make smaller in the future. This, and the `main()` function is basically an extension of the original function as it does everything that occurred prior. 

Still quite a bit of work to do, mainly I ought to focus more on Typescript and smaller modules as I mentioned. 

Otherwise, to silence the linter I had to capitalise the class and change all corresponding references in other files. 

Hope it all works the same, made sure that tests all pass, though I had to silence the linter in a few test files because it did not like me doing: `new Sortable(...)`, ie. using it as a side effect.

Thanks!


lukasoppermann:
Hey @jMuzsik,

thanks for this work. However I will not merge this. While I do really want to turn this package into a class (see branch [class](https://github.com/lukasoppermann/html5sortable/tree/class)) I want to do it differently.

My current plan is to clean up everything and turning the package into a class only when it can be done very cleanly and easily. There is much work to be done before this is possible.

For example the whole data situation needs to be solve, which is what I am currently working on.
Also all functions need to be cleaned up and condensed. And the dependencies on global variables (with the exception of the `store`) need to be removed. You basically had to move everything into the class, because of the way the package is currently structured. This makes the class very overwhelming. The class should mostly delegate to small modules so it is much easier to understand.

I hope you understand my reasoning, if not, please do let me know and ask any questions you have.


**Also:** could you please send a PR with only your changes from `src/types/main.d.ts` and `tsconfig.json`. This can be easily merged. If you want I would be very happy to discuss aspects where you can help extract and improve functions of the current code to move us more towards being able to create a good and simple class.

Thanks.

jmuzsik:
@lukasoppermann ok I understand! Either way I see most of what I am doing as an experience rather then some sort of expectation driven PR's. 

I'd like you to point the way as to what you'd like me to do then, as you mentioned, seeing as you have a very clear picture of what to do as, quite obviously, it's impossible for me to have as clear a concept as to what needs to be done then you do. I'm open to whatever you recommend, obviously I am more junior a coder but with some time I can figure out most tasks.

lukasoppermann:
Hey, I will write you some options of what would be helpful tomorrow and you can see what is interesting for you. Also I would really appreciate the two things I mentioned in a PR.  

It is always complicated to see a vision from the outside. Also my not wanting to move into the class yet has nothing to do with your code, but just that the package's current code is not refactored enough yet to transfer it into a clean class. 

Either way I am very happy to have you contributing. 👍

lukasoppermann:
Hey @jMuzsik, the following things would be very interesting for the project:

- fixing #286 to add touch events
- adding a `contain` option to keep items within a sortable container: #276
- investigating #384, documenting the findings and hopefully removing the fix
- making buttons & links easily draggable #341
- remove items when dropped outside valid target #400 this would also mean allowing for a second option: #313
- adding correct cursor when interacting with items: #351

Just let me know which one you are interested in and we can discuss the approach in more detail in the specific issue. 😃

jmuzsik:
@lukasoppermann Definitely appreciate this project, I am quite new to programming and am switching careers towards it soon so this experience is invaluable. 

I'm gonna have a go on adding touch events. I just started reading about it and it appears to be something that, while it will be quite a work to implement would nonetheless be very interesting work to do.

I'd defintely need to do much research and get a better understanding of how the library works prior though. I'll being working on it today but it may be some time before I have anything that works properly but I will push `[WIP]` branches here and there.

Thanks!

lukasoppermann:
Hey @jMuzsik,
I am very happy that you are using this project to improve your skills. This is a really cool thing, as you get experience and also benefit the community by improving this project. 👍 

Working on touch event issue sounds very good. If you have any questions please just asked them in the `[WIP]` PR and I will try to help out. Feel free to push a non-working PR as well. Looking forward to it. Cheers!

Issue #388, opened 2018-03-20
Hover on another element after dragstart


lukasoppermann:
When hover is active and `dragstart` fires, if an element moves into the position of the hovered item it gets the hover status.

Maybe on dragstart hover should be removed.

mfeherpataky:
I can look into this one, as it is on my TO-DO list (it is somewhat confusing for some users that notice it), but only after #394.

lukasoppermann:
That would be great. 👍

PR #402, opened 2018-03-29
Fix/388 hover on another element after dragstart

mfeherpataky:
Solves #388, to the extent that `options.hoverClass` will not be added to any new item for as long as dragging takes place. 

Chrome still maintains pseudo-class :hover at the original location of the item that is being drag. 
Recent Firefox and Safari, however, do not maintain CSS :hover pseudo-class at the original item location. 
The `README.md` warning not to use CSS :hover selectors for styling still holds true.


lukasoppermann:
Geniously simple fix, thanks a lot. 👍

PR #431, opened 2018-05-15
Fixed IE bug , added a polyfile to 'matches' method in Element class

av-nexter:


lukasoppermann:
Hey, sorry but this was previously removed on purpose to slim down the code. You can read more about the decision in the readme: https://github.com/lukasoppermann/html5sortable#polyfills-facing-towards-the-future-instead-of-the-past

PR #451, opened 2018-08-16
Use the event target's ID for data for compatibility with Firefox

### LINKED ISSUE #450: FireFox occasionally redirects to "arbitrary.com"
### I'm opening this issue because:
After dragging and dropping table rows, FireFox triggers a search for "arbitrary", which automatically redirects to arbitrary.com

### supporting information:
- I use sortable version: 0.9.3
- I tested in the following browser (incl. versions): FireFox 61.0.1 64 bit 
- Windows, OS X/macOS, or Linux?:### How can the issue be reproduce the problem? Windows 10, MacOS
Please add a complete description of how to reproduce the problem.

I've got a table of data and making the rows sortable. In FireFox, after about 10 drag and drops, FireFox immediately triggers a search for "arbitrary" and automatically redirects to arbitrary.com

I believe this issue is related to line 458 of html5sortable.js:
event.dataTransfer.setData('text/plain', 'arbitrary');

ericparshall:
fixes #450

### DIFF
--- a/__tests__/setDragImage.test.ts
+++ b/__tests__/setDragImage.test.ts
@@ -40,6 +40,12 @@ describe('Testing setDragImage', () => {
     get () {
       return this
     }
+
+    get target ():any {
+      return {
+        id: 1
+      }
+    }
   }
 
   test('No event provided', () => {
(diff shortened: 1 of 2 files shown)


lukasoppermann:
Looks good, thanks. 👍

PR #470, opened 2018-11-02
Later calls to `sortable()` mustn't overwrite options set in earlier calls

schuetzm:
When calling `sortable()` to enable sorting for newly added DOM elements, the options that were set by the first call are reverted to their defaults. This contradicts the [documentation](https://github.com/lukasoppermann/html5sortable#reload).

This PR fixes the problem. I've also added a test case.

Note that I'm not sure about the relationship between `store(...).config` and `_data(..., 'opts')`. AFAICS, the latter is only read from, so I assume the change in line 258 is safe.

### DIFF
--- a/__tests__/options.test.ts
+++ b/__tests__/options.test.ts
@@ -33,4 +33,28 @@ describe('Test options from sortable', () => {
     // also test a default value to check if they stay the same
     expect(sortableElement.h5s.data.opts.draggingClass).toEqual('sortable-dragging')
   })
+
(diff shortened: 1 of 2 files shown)


lukasoppermann:
Hey @schuetzm this looks pretty good. I will need to check it and merge it soon. Thank you. 👍 Would be awesome if you have more PRs to help.

lukasoppermann:
Merged 👍

PR #492, opened 2019-03-01
make sure sortenter event is fired when a dragged element enters container the first time

christophe-g:
Motivation: 
- Be able to listen to `sortenter` event - useful when dragging to a different list and wish to update placeholder content.
- Make sure we have a way to trace back the original target in `sortstart` and `sortenter` event. When sortable is used with copy = true, the dragged item is a deep-copy of the original target. This prevents accessing the real original target. This is a problem in some cases (like in polymer dom-repeat component) where only the original target holds some linked properties.


christophe-g:
Sorry for the number of commits on this one. Had to reconcile history tree between my branch and yours.

christophe-g:
@lukasoppermann - hey 
If you agree with the proposed changes, I'll re-push with a clean commit history (just re-branch from your master and replay the changes).

christophe-g:
Closing as replaced by #494

PR #517, opened 2019-10-21
Append placeholder to non-empty containers. Fixes #172

### LINKED ISSUE #172: Height of droppable area changes after first element dropped
Here's a jsfiddle that shows an example of what I mean:

http://jsfiddle.net/6y9w4opk/

If you drag one item from the bottom list into the first, you can drag it anywhere over the `<ul>` and the item will be moved over correctly (you can see it by the dashed outline of the placeholder).

When you try to drag a second item into the top list, you have to drag it over the previous `<li>` element or it won't allow you to drop the new item, even though the `<ul>`'s height hasn't changed.

Here's a video showing what I mean as well:

https://youtu.be/I39e2nlj2HY

It's not that anything breaks, but this has caused confusion for users in two separate test sessions, so we'd like a way to keep the entire `<ul>` area droppable and have the dropped `<li>` end up at the end of that list.

Please let me know if there's a workaround or fix I can try. Thanks!

erickoledadevrel:
Previously the code was only allowing appends when the sortable container was empty. There doesn't seem to be a reason for this, and it prevented users from dropping items in the empty space of a tall, non-empty container.

### DIFF
--- a/src/html5sortable.ts
+++ b/src/html5sortable.ts
@@ -570,7 +570,7 @@ export default function sortable (sortableElements, options: configuration|objec
             return data.placeholder
           })
         // check if element is not in placeholders
-        if (placeholders.indexOf(element) === -1 && sortableElement === element && !_filter(element.children, options.items).length) {
+        if (placeholders.indexOf(element) === -1 && sortableElement === element) {
           placeholders.forEach((element) => element.remove())
           element.appendChild(store(sortableElement).placeholder)
         }


erickoledadevrel:
I looked into adding tests, but it seems like a lot of new ground to break (given that this function is not under test at the moment) and I'm not comfortable enough with the codebase nor do I have the time to commit at the moment.

I also tried to extract these lines (as you mentioned) to make it easier to put them under test, but I don't think it refactors very naturally. The function name would have to be something like `maybeAppendPlaceholder` and it didn't smell like a good place to break up the code.

lukasoppermann:
Thanks, I'll look into it and merge it. Did you do some manual testing?

erickoledadevrel:
Yes, I built the js library and swapped it into my app. Seems to work well, but I don't know if there were some edge cases this code was initially defending against.

lukasoppermann:
Hey, so there is an issue that is created with this PR. If you try to add an item between two items (so that the cursor is neither on the item before nor the item after where you want to insert) it now inserts at the end of the list. Could you maybe add a for for this?

Sorry for closing the PR, please just send a new one (you can start by resending this and if you can sending an update later, I will leave it open for the time being)

erickoledadevrel:
Ah, I see. Before it would just do nothing in that case?

lukasoppermann:
Yes, it does nothing. Which is "better". Of course it would be ideal if it would check if there is something around and insert it correctly, but this is quite some work. I think nothing would Generally be fine. So we would need to check if there is an item above the cursor if not, insert it at the top of the list. If there is one above but none below, insert it at the end.

Does that make sense?

PR #538, opened 2020-04-27
Minor update to README

Barabazs:
- Correct mistake concerning "e.detail.destination.container"
- Emphasise that connectWith is deprecated in section about acceptFrom


Barabazs:
I would also like to point out that the readme still includes multiple references to the deprecated function connectWith. I am unsure if those references can simply be changed to reference acceptFrom. 

Maybe you could have a look and update where it is appropriate. Or I'd be glad to update those references if you can confirm that the methods are interchangeable.

lukasoppermann:
It would be great if you could update the rest as well. I am currently not sure if this does apply to acceptFrom as well. Would you have a moment to test it?

PR #543, opened 2020-07-01
Updated Documentation of acceptFrom

### LINKED ISSUE #520: Sortable freezes when acceptFrom list does not exist
Originates from #519 

## Current behavior:
When `acceptFrom` is used, but no sortable with the selector can be found the sortable freezes.
**Needs validation:** This should be validated with an example on codepen.

**Desired behavior:**
1. Catch error from [isConnected](https://github.com/lukasoppermann/html5sortable/blob/master/src/isConnected.ts) and log to console.
2. Make sure the sortables work as if `acceptFrom` was not present

kaffarell:
resolves #520

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 ![](https://repository-images.githubusercontent.com/275870366/15880680-ba44-11ea-888b-8211b9431f94)
 
-<h1 align="left"> HTML5Sortable </h1>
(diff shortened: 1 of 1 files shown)


kaffarell:
@lukasoppermann can you please look into this

lukasoppermann (approved):

PR #544, opened 2020-07-01
Fixed box-sizing height calculation

### LINKED ISSUE #432: elementHeight does not take in to account the box-sizing method
the code used to fix the height of the placeholder does not take in to account the box-sizing method - and always adds top/bottom padding to computed height. When an element has box-sizing: border-box, the padding is already included in the computed height.

I use sortable version: 0.9.2

### How can the issue be reproduce the problem?
 - ensure the sortable items have top/bottom padding and box-sizing: border-box
 - enable the forcePlaceholderSize option
 - sort the items

kaffarell:
Fixes #432
Checks if box-sizing is set to border-box, if it is, return only height, otherwise height with padding added.
Wrote tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 ![](https://repository-images.githubusercontent.com/275870366/15880680-ba44-11ea-888b-8211b9431f94)
 
-<h1 align="left"> HTML5Sortable </h1>
+<h1> HTML5Sortable </h1>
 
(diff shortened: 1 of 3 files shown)


kaffarell:
@lukasoppermann on the other tests of elementHeight I also changed it so that the div is not inserted into the DOM but it gets calculated as JS element, is that alright? (The previous tests also didn't insert the element in the DOM)

lukasoppermann (approved):

lukasoppermann:
Nice one 👍

kaffarell:
@lukasoppermann did I break the Sortable Grid on the example website?
Was it previously also this buggy?
Will look into it today

lukasoppermann:
Let me check, but I don't think this is possible, as I did not create a new build yet.

The grid is not working great yet.

kaffarell:
@lukasoppermann Can you please reopen this pr?

kaffarell:
OK 👍  noticed it now

PR #546, opened 2020-07-03
Inserted Table of Contents in readme

kaffarell:
With the TOC you can navigate around the readme more easily

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,77 +18,36 @@
   * [Looking for Co-Maintainer](#looking-for-co-maintainer)
   * [Features](#features)
   * [Framework adapters](#framework-adapters)
-
 - [Installation](#installation)
-
 - [Examples](#examples)
-
 - [Docs](#docs)
-
   * [Usage](#usage)
-
   * [Styling](#styling)
-
   * [Nesting](#nesting)
-
   * [Events](#events)
-
-    + [sortstart](#sortstart)
-    + [sortstop](#sortstop)
-    + [sortupdate](#sortupdate)
-
   * [Options](#options)
-
-    + [items](#items)
-    + [handle](#handle)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,20 +1,94 @@
...
 
 > **Lightweight vanillajs micro-library for creating sortable lists and grids using native HTML5 drag and drop API.**
 
-## Community maintained
+
+
+# Table of Contents
```
lukasoppermann, line 15: Sorry for the trouble, but those should all stay h2s. From a web hierarchy point of view we don't want too many h1s. (Optimally just one or two.)

README.md
```diff
@@ -1,20 +1,94 @@
...
+
+  * [Community maintained](#community-maintained)
+  * [Looking for Co-Maintainer](#looking-for-co-maintainer)
+  * [Features](#features)
+  * [Framework adapters](#framework-adapters)
+
```
lukasoppermann: We should get rid of the whitespace between the list items, this makes the page too long.

README.md
```diff
@@ -1,20 +1,94 @@
...
+  * [Usage](#usage)
+
+  * [Styling](#styling)
+
+  * [Nesting](#nesting)
+
+  * [Events](#events)
+
+    + [sortstart](#sortstart)
```
lukasoppermann: I feel we can drop the subitems. Events, Options, Methods, Benefits, Known Issues are enough, the subpoints don't need to be here. Otherwise the list gets too long.

lukasoppermann (changes requested):

kaffarell:
:+1:  Fixed know, thanks for the feedback!

lukasoppermann (approved):

lukasoppermann:
Great, thank you.

PR #557, opened 2020-08-18
Renamed index.ts to getIndex.ts 

### LINKED ISSUE #556: Rename index.ts to to getIndex.ts
`index` is problematic because of npm using index as default files

- [ ] rename file to `getIndex.ts`
- [ ] import function as `getIndex` so that when it is used it is used as `getIndex`

kaffarell:
FIxes #556


lukasoppermann (changes requested):
Hey @kaffarell this is great and very quick. To make it even easier to read please also rename the imports from index to get index as well so that the function that is used is always getIndex. Thanks 👍

kaffarell:
Changed it, by the way what is the underscore at the beginning of the imported function meaning?

lukasoppermann:
Hey @kaffarell thank you. 👍 
The underscore is just a naming convention to denote functions or methods that are internal and should not be used as they are not part of the public api and thus may change. 
However, here it is a remanence of much earlier versions. Could you please update it here to NOT include it (I know it was there already, but it was good of you to spot it and we should clean it up), thanks.

Afterwards I will be able to merge it. 😄

kaffarell:
@lukasoppermann  This should be complete

lukasoppermann (approved):

lukasoppermann:
Awesome, @kaffarell I was actually only referring to the _index. method, but that is great. :D

PR #562, opened 2020-10-12
#561 replace travis with GitHub action

### LINKED ISSUE #561: Replace Travis with github action
To simplify the setup replacing the tests on travis with a github action would be great.

Gontrum:
Hi everyone,

I'm not 100% sure if it fits like this. Actually I wanted to resolve #561 . Maybe you have some tips for me:

- Unfortunately the build fails with `npm run standard`, but this happens in `master` too. Should I just ignore that?
- I removed `greenkeeper-lockfile` from the action, because it says here https://github.com/greenkeeperio/greenkeeper-lockfile that this is now build-in. 

Does it fit otherwise? 
Best regards


.github/workflows/html5sortable.yml
```diff
@@ -0,0 +1,29 @@
...
+        key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+        restore-keys: |
+          ${{ runner.os }}-build-${{ env.cache-name }}-
+          ${{ runner.os }}-build-
+          ${{ runner.os }}-
+
+    - name: Build and test
+      uses: actions/setup-node@v1
+      with:
+        node-version: 12.x
```
atodorov: From the last Travis CI log:

```
$ node --version
v14.13.1
```

atodorov (changes requested):
Overall looks good to me but not sure why the CI job isn't starting.  

@Gontrum do you have the job running on your own fork ? If so can you post a link to the logs.

.github/workflows/html5sortable.yml
Gontrum: sure, @atodorov. You should be able to see it here: https://github.com/Gontrum/html5sortable/runs/1242446082?check_suite_focus=true

I also updated the job to use node version 14.x. Thank you for your feedback.

lukasoppermann: The failing test are happening since I updated some of the dependencies. Somehow though not locally, any ideas why they only happen on travis / github actions?

Gontrum: When I run `npm run standard` locally, the error occurs too. 
Do you have some global dependencies installed, which could overwrite local dependencies?

lukasoppermann: Good idea, but somehow not, I uninstalled all global npm modules and still not. And I just cloned the repo.

Gontrum:
I changed the version declaration to a string. Otherwise the actions always takes node version 10 ;-) 
Sorry for that!

.github/workflows/html5sortable.yml
Gontrum: I could give it a try inside a docker-environment, later this day. Maybe we can narrow down the problem this way :-)

Gontrum: I got it running, now. Your dependencies were fine but something went wrong with coveralls and travis. But that doesn't matter now, because a different mechanism is used for github. 
Standard failed because of some globals which were not part of the standard configuration. I also added a comment for disabling eslints `no-use-before-define` inside `store.ts` and hope that's fine with you. 

i also had to adjust the readme, because the travis-badge is of course out of date after the update 🙂

Gontrum: @atodorov I just realized, that the action is still not running with the pull request. 
I think this is because the workflow is not yet part of the main-branch. 
Since the master build with travis is currently broken and the [github-build in my fork works](https://github.com/Gontrum/html5sortable/runs/1243550206?check_suite_focus=true): do we want to merge and check if it works here?

lukasoppermann:
Hey @atodorov this should satisfy your change request so we can merge it, correct?

atodorov (approved):
LGTM now

atodorov:
@Gontrum can you rebase on top of the latest master and resolve the conflicts please.

Gontrum:
> @Gontrum can you rebase on top of the latest master and resolve the conflicts please.

Sure :-)
Should be clean, now. Thank you guys for your help.

atodorov:
@lukasoppermann I don't see any obvious conflicts here but I can't "Rebase & merge". The only available options for me are "Merge pull request (with a merge commit)" or "Squash and merge".

lukasoppermann:
@atodorov same here. I always "squash & merge" in any case, as it keeps the history a bit more clean & understandable. Does anything speak against squash & merge?

atodorov:
> @atodorov same here. I always "squash & merge" in any case, as it keeps the history a bit more clean & understandable. Does anything speak against squash & merge?

FTR: I think the issue here was a merge commit in the PR itself. 

I personally prefer "Rebase & merge" to keep the history clear and also keep the individual commits separate so we can trace back history more granularly. In this particular case I agree squash was the right thing to do since there were many commits fixing items from the code review process or adding changes missing from the original commit.

Gontrum:
@atodorov thanks for the hint with the merge commit, I will keep that in mind.

lukasoppermann:
@atodorov so if a PR has multiple commits that make sense by themselves I agree with the "Rebase & merge". Often though the PR history is not cleaned up. But we could ask for a clean history as well. 👍

Issue #445, opened 2018-06-28
Not working on IE 11


Polm90:
I''ve tried the demo https://lukasoppermann.github.io/html5sortable/index.html on IE11 (on Windows 10 with April Update) and nothing works. But in the console there are no errors.

Any help?

Thank you

lukasoppermann:
Hey @Polm90 sadly I have no windows machine available to easily debug this and I am pretty short on time. If you can help with figuring out what the problem could be and maybe even how to fix it, that would be great.

Any more info you can provide would be great.

Polm90:
Unfortunately it's a strange behaviour because the browser gives no errors (or warnings - in console). But when you try to drag items as you do on other browsers, the browser selects/highlights the element (as you select some text) and there is no possibility to drag.

I didn't have time to try to figure out why, but if I'll find anything new I'll keep you update.

jmartsch:
There are JavaScript errors in the console which I think are a result of using arrow functions, which are not supported in IE.
`itemSerializer: (item, container) => {
                item.parent = '[parentNode]'
                item.node = '[Node]'
                item.html = item.html.replace('<','&lt;')
                return item
              },
![iexplore_2018-07-21_02-51-01](https://user-images.githubusercontent.com/551551/43030482-eef1cd54-8c90-11e8-9c13-2a5e250c85a1.png)
`

lukasoppermann:
Hey, there should not be any arrow functions in the code, as the code is converted which removes those error functions.

jmartsch:
Please take a look at the source code of the demo page @ https://lukasoppermann.github.io/html5sortable/index.html. In line 74 there is clearly an  arrow function. And on line 80 and 85 also.

lukasoppermann:
🤦‍♂️  I thought you were talking about the actual code of the sortable. Of course you are right here. 

I remove them now, so does this fix the errors?

jmartsch:
The error is gone, but other errors still excist:
The object does not support the property or method "entries"
This error occurs five times on the page.
![iexplore_2018-07-21_11-38-41](https://user-images.githubusercontent.com/551551/43034322-a3a1f10e-8cda-11e8-8871-1fefd1a13d5f.png)

And on line 502 you make use of template strings, which are also not supported in IE.

lukasoppermann:
Hey, could you maybe send a PR? I am out and I also have a Mac so I can't test on IE. That would be great. Thanks

kretasei:
[...]

PR #568, opened 2020-10-14
[WIP] Update rollup-plugin-typescript to rollup-plugin-typescript2

### LINKED ISSUE #560: The dependecy 'rollup-plugin-typescript' is deprecated
we should change it to [this](https://www.npmjs.com/package/@rollup/plugin-typescript) '@rollup/plugin-typescript'. Don't know if this would brake anything.

kaffarell:
Will fix #560

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8461,9 +8461,9 @@
       }
     },
     "rollup": {
-      "version": "2.29.0",
-      "resolved": "https://registry.npmjs.org/rollup/-/rollup-2.29.0.tgz",
-      "integrity": "sha512-gtU0sjxMpsVlpuAf4QXienPmUAhd6Kc7owQ4f5lypoxBW18fw2UNYZ4NssLGsri6WhUZkE/Ts3EMRebN+gNLiQ==",
+      "version": "2.30.0",
+      "resolved": "https://registry.npmjs.org/rollup/-/rollup-2.30.0.tgz",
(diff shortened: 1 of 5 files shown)


atodorov:
> Will fix #560

Can you add this to commit log and also rebase b/c the latest master branch updated a bunch of other dependencies.

Otherwise LGTM so not sure why the subject says "WIP".

kaffarell:
Ok, have done the changes. I included WIP in the title because there are a lot of  other error down the line. I removed the IE fix but I cannot check if it is used because IE11 is still not working ( #445 ). The problems are typescript errors, every usage of options is using the configuration type but it also has other types.
https://github.com/lukasoppermann/html5sortable/blob/a19d77834cf63eccb8d1fba12b6cb27f6be1d464/src/html5sortable.ts#L230-L237
So I could add a `<configuration>` before every usage of options but I don't think this is the best solution. I haven't found out when the object string or undefined type on options is used, so we could maybe remove them and write:
`options: configuration`

kaffarell:
All the errors look like this:
```
semantic error TS2339: Property 'items' does not exist on type 'string | void | configuration | HTMLElement'.
  Property 'items' does not exist on type 'string'.
```

lukasoppermann:
Hey @kaffarell please verify what I am saying, but I think you could cast options to <configuration> here: https://github.com/lukasoppermann/html5sortable/blob/a19d77834cf63eccb8d1fba12b6cb27f6be1d464/src/html5sortable.ts#L267 because at this point it will always be a configuration, correct?

Otherwise we would probably need to add a function that deals with the options parameter and returns a configuration, but try the above mentioned first please.

lukasoppermann:
@kaffarell any news so this can be made mergable?

Issue #456, opened 2018-09-04
On fast drops the drop event fires before placeholder is created and component breaks


breugen:
### supporting information:
- I use https://lukasoppermann.github.io/html5sortable/index.html

### How can the issue be reproduce the problem?
Just drag as fast as you can many elements from a list to another list like in the picture. Usually after 10 tries you should be able to break it.
Real users reported this, probably on slower computers it happens even more, no need to stress the app.

![evidence](https://user-images.githubusercontent.com/2683127/45028940-9d22ab00-b04f-11e8-8c38-4c7a62fd4f7d.PNG)
![more-evidence](https://user-images.githubusercontent.com/2683127/45028949-a0b63200-b04f-11e8-8a77-700b977f2578.PNG)

lukasoppermann:
Thank you for reporting, do you already have a guess as to how this can be fixed?

breugen:
No, sorry, I just temporary patched it by checking if the placeholder exists, and if not I return and delete any clone (display none) from the source list . The user perceives this as 'did not worked' but at least can try again and the app does not break. It's a bit strange for me the '_debounce' code, that seem to allow drop before rendering the placeholder. But have not dug deep enough into the code to fully understand it in order to mess with that .

lukasoppermann:
Hey @ebreaur,

as I am not using this plugin myself anymore (and also not working as a developer) I am not really in the position to dedicate time to solve this issue. However if you would like to fix it I am happy to support you as good as I can.

Did you try disabling the `_debounce` to see if this is the issue?

PR #677, opened 2021-03-03
Potential change to address Issue#456

Lemonsavery:
https://github.com/lukasoppermann/html5sortable/issues/456

### DIFF
--- a/docs/html5sortable.js
+++ b/docs/html5sortable.js
@@ -1115,9 +1115,11 @@ var sortable = (function () {
                   // only elements in DOM
                   .filter(isInDom)[0];
               // attach element after placeholder
-              insertAfter(visiblePlaceholder, dragging);
-              // remove placeholder from dom
-              visiblePlaceholder.remove();
+              if (visiblePlaceholder && visiblePlaceholder.nextElementSibling != null) {
+                insertAfter(visiblePlaceholder, dragging);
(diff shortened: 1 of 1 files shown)


Lemonsavery:
I encountered the same issue as described in #456, and this was part of my fix for it. My issue would occur while using Chrome, but not while using Edge.

lukasoppermann:
Hey @Lemonsavery 
thank you very much for the PR. I will have to look into it to to make sure it has not repercussions and hope to be able to merge it very soon. Great. 👍

lukasoppermann:
Hey @Lemonsavery can you please explain the issue it is solving again? I have the feeling the issue mentioned in the ticket persists, but this may be due to me not really getting how to reproduce it. 😁

Lemonsavery:
@lukasoppermann Here are two gifs of rows being dragged and dropped quickly/normal. Drag events are printed to Chrome DevTools console. When dragged and dropped too quickly, the 'drop' event can occur before the placeholder row is in place. insertNode() then tries to use the placeholder as referenceNode, but referenceNode.nextElementSibling is null, **so the dragged row vanishes**. Then, later, the placeholder row is created.
| buggy drag | correct drag |
|:---:|:---:|

lukasoppermann (approved):

PR #773, opened 2021-06-27
fixing a bug where dragging.style.display was not restored on dragend

k2tomasz:


lukasoppermann:
Hey @tkrakowiak,

thanks for the fix.

Can you please rebase and only commit `src/html5sortable.ts` I will create the rest automatically in the release.

Thanks

Issue #424, opened 2018-05-07
Items disappear on drag


CWSites:
### I'm opening this issue because:
When I drag an item to any area inside or outside of the wrapper it disappears from the screen. There are no console errors. This may be related to #278 although I'm not on an Android device.

### supporting information:
- I use sortable version: 0.9.2
- Google Chrome 66.0.3359.139
- macOS

### How can the issue be reproduce the problem?
Click and drag any item, anywhere on the screen.

See screen recording here - https://www.dropbox.com/s/7i91kx4i9ercaww/html5sortable.mov?dl=0

### My Code
```
import sortable from '../../../node_modules/html5sortable/dist/html5sortable.es.js';

sortable('.sortable', {
    forcePlaceholderSize: true,
    placeholderClass: 'sortable-placeholder',
    hoverClass: 'is-hovered'
});

<Row className="sortable">
  <Col lg={ 4 }>
    <Panel>...</Panel>
  </Col>
  <Col lg={ 4 }>
    <Panel>...</Panel>
  </Col>
  <Col lg={ 4 }>
    <Panel>...</Panel>
  </Col>
  <Col lg={ 4 }>
    <Panel>...</Panel>
  </Col>
  <Col lg={ 4 }>
    <Panel>...</Panel>
  </Col>
  <Col lg={ 4 }>
    <Panel>...</Panel>
  </Col>
</Row>

lukasoppermann:
Hey, I have no idea why. I would guess the item is removed because it loses its original parent and is dropped on an invalid target. But this is a wild guess. Can you maybe provide a live version so that it can be tested?

If you find any more infos that might be related, please add them here.

CWSites:
It must be something to do with how I'm using it because I have the same configuration as my codepen (https://codepen.io/CWSites/pen/wjeNbb) and it doesn't have issues.

I'm doing this on a React App and I have other libraries involved as well but I'm not certain where the underlying issue is. 

I sent you a collaborator invite to a private repo of my code. It's happening for me when I'm running it locally so you shouldn't have to worry about setting up a server to run the app.

CWSites:
Hey @lukasoppermann, I updated my code base a bit and I'm unable to reproduce this. Feel free to close it out as an edge case.

lukasoppermann:
Nice

ammmir:
[...]

Issue #278, opened 2017-08-22
Dragging to far on Android tablet makes item disappear


mamcgree:
Greetings..

Your sortable JS works great on desktop browser.  But if you use an android tablet and you drag items to far they disappear.  Not sure if this is by design or not.

Thank you in advance for any advice you my offer.

Cheers

Mark..

lukasoppermann:
Hey, this is definitely not by design. 😉  Anyway if you can provide any information or even better a fix that would be very appreciated.

jneb:
It is reproducable on Edge (windows 10). Just take the example file, take the first item of the first list, drag far enough up, and poof! Vanished.

jneb:
This appears to be the same bug as #424

Issue #163, opened 2015-07-22
IE10/11: User is able to remove elements from the page after highlighting and dragging


hnguyen08:
In IE 10/11 only (tested IE10 and IE11, not occurring in Chrome or Firefox or IE9 surprisingly):
1. Visit the [demo page](http://lukasoppermann.github.io/html5sortable/examples/index.html)
2. Scroll down to the section with the handles under "Connected Lists"
3. Highlight a few elements of the list.  In the picture below, I highlighted the subitems of Item 2 through Item 5
4. Use the handle to drag one of the highlighted elements to another part of the list.  In the picture below, I dragged Item 5 to above Item 2
5. See that the elements have disappeared

Step 3:
![image](https://cloud.githubusercontent.com/assets/2112727/8830578/ee28d51e-306a-11e5-9c3c-e3557b5419a3.png)

Step 4:
![image](https://cloud.githubusercontent.com/assets/2112727/8830586/fce74ce8-306a-11e5-8110-32a804bac5d3.png)

ncareau:
Confirmed on IE11   11.63.10586.0

Any quick fix on this issue ?

iArnoldo:
Anything new to this issue?

lukasoppermann:
Hey, I do not have any IE10/11 available for testing.

Are there any errors in the console? @iArnoldo ?

Can anybody pin-point the bug and send a PR to fix it?

jneb:
Probably the same bug as #424

PR #841, opened 2021-11-05
Bug fix for issues #424, #278, #163

jneb:
Disclaimer: I am only playing with this for an hour or so.
But this fix makes complete sense: the effect is that if an item is not dropped, there's no code that makes the dragged object visible again, making it disappear.
All I did was make sure that it is made visible at dragend regardless of the drop.
But I do think that I just fixed no less than three issues: #424 and two issues that appear to be the same (#278 and #163).
But I tested on only one machine.

### DIFF
--- a/src/html5sortable.ts
+++ b/src/html5sortable.ts
@@ -421,6 +421,7 @@ export default function sortable (sortableElements, options: configuration|objec
 
       dragging.classList.remove(options.draggingClass)
       attr(dragging, 'aria-grabbed', 'false')
+      dragging.style.display = ""
 
       if (dragging.getAttribute('aria-copied') === 'true' && data(dragging, 'dropped') !== 'true') {
         dragging.remove()


jneb:
See my comment at issue #424. The fix is almost right, but not quite.

jneb (commented):
That probably should rather be
`dragging.style.display = dragging.oldDisplay
      delete dragging.oldDisplay
`
considering the changes for issue #777

jneb (commented):
It appears that #777 fixes the issue, but this change is not visible in the current vversion.
In particular, it is the lines from 428:
`      if (dragging.oldDisplay !== undefined) {
        delete dragging.oldDisplay
      }
`
but they do not appear in the demo example.

lukasoppermann:
Hey @jneb what is your latest take on this after investigating? Does this still fix the bugs?

Also please change it to single quotes to appease the gods of linting. Thanks.

jneb:
I had a good look at the git blame, and found out that a fix you did for
months ago fixes it. ( It does for me, at least. )
And from my analysis of what I think the code is supposed to do, combined
with what I think it's the cause of the problem, your fix (#777) should
handle all cases.
I checked it against Chrome, Edge and Webview; those are all I have.
My conclusion is that the main problem is that the library code in the demo
directory isn't up to date.
But it didn't really figure out which is the latest "official" version;  I
was confused to be honest. I guess this confusion may also cause big
reports from others.

But thanks for a great product!
I use it for a personal planning program that I am developing. It started
just for myself (I am a terrible and messy planner, and I do need help :-)
). The goal is to publish it after making it useful for outsiders.
-Jurjen



Op za 6 nov. 2021 15:22 schreef Lukas Oppermann ***@***.***>:

lukasoppermann:
Hey, @jneb can you please fix the linting error so I can go ahead and merge it?

jneb:
I'll have a look.
- Jurjen

Op vr 12 nov. 2021 09:04 schreef Lukas Oppermann ***@***.***>:

lukasoppermann:
Awesome, it's just a tiny change, but it blocks the merge. 😉

jneb:
If I have a good look at the version of code this fix is merging with, it shouldn't be necessary to change anything at all!
There is code a few lines lower that looks like it does what my code was intended to do.
I'll do a final check but I am already almost convinced the current master branch is OK.

jneb:
Yes, I checked it. The problems are caused by the "docs" directory containing an old version of the library, and some people (like me) being confused by that version.
Sorry for creating confusion and misinformation.

PR #854, opened 2021-12-20
Add keep css class option 

heyyo-droid:
Allow to keep some css class from dragged element and apply them to placehoder.


src/html5sortable.ts
```diff
@@ -482,6 +482,10 @@ export default function sortable (sortableElements, options: configuration|objec
           dragging.style.display = dragging.oldDisplay
           delete dragging.oldDisplay
         }
+        // We need to reset placeholder if any css class were added.
+        options.keepCssClass.forEach(className => {
+          store(sortableElement).placeholder.classList.remove(className)
```
lukasoppermann, line 487: This can be done in one step https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/remove sot that all classes are provided at once.

I think this should work: `remove(...options.keepCssClass)`

src/html5sortable.ts
```diff
@@ -552,6 +556,13 @@ export default function sortable (sortableElements, options: configuration|objec
         store(sortableElement).placeholder.style.height = draggingHeight + 'px'
         store(sortableElement).placeholder.style.width = draggingWidth + 'px'
       }
+      // Keep some CSS class fo the dragged element, to keep style.
```
lukasoppermann, line 559: fo = for in comment

src/html5sortable.ts
```diff
@@ -552,6 +556,13 @@ export default function sortable (sortableElements, options: configuration|objec
...
+      // Usefull if sortable elements have different dimensions according class.
+      options.keepCssClass.forEach(className => {
```
lukasoppermann, line 561: I think from a performance point of view you should first get the elements you want to add. Maybe use filter to get the elements that are in both arrays.

Afterwards just like with remove, you can `add` all at once. DOM interactions are often expensive.
`add(...comparedArray)`

src/types/configuration.d.ts
```diff
@@ -1,7 +1,8 @@
 interface configuration {
   items?: string,
   handle?: string,
-  forcePlaceholderSize?:boolean;
+  forcePlaceholderSize?: boolean,
+  keepCssClass?: Array<string>,
```
lukasoppermann, line 5: Let's rename this to `keepClassesForPlaceholder`.

lukasoppermann (changes requested):
Please remove all files in the `dist` folder from the pr. This is always done on release only.

lukasoppermann:
Hey @heyyo-droid thank you for the PR. I left a few comments.

I am thinking if it may make more sense to have an initialize placeholder function that the user can define.
This would enable this use case but allow for more flexibility.

We would add a `initPlaceholder: fn` option. This would get the element it is a placeholder for and a clone of the placeholder and would return the placeholder element which in turn would be attached to the dom and moved into position.

What do you think?

src/html5sortable.ts
heyyo-droid: Nice, I will change that.

src/html5sortable.ts
heyyo-droid: Sorry you lost me in the first sentence.
Second sentence nice did't know that too.

